PR #30, opened 2020-07-26
Add basket handle cutouts and optimize shape map

ChloeDawn:
- Removed solid handle pixels from texture and added elements to the model with inverted normals to fill the spacing
- Updated the block's voxel shapes and adjusted the model's face culling to prevent x-ray
- Replaced the mutable hash map of directional shapes with an immutable enum map for better lookup performance

<p align="center">
  Basket model render
</p>


src/main/java/vectorwing/farmersdelight/blocks/BasketBlock.java
```diff
@@ -26,26 +28,58 @@
...
-		Map<Direction, VoxelShape> result = new HashMap<>();
-		result.put(Direction.DOWN, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(2.0D, 0.0D, 2.0D, 14.0D, 14.0D, 14.0D), IBooleanFunction.ONLY_FIRST));
-		result.put(Direction.UP, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(2.0D, 2.0D, 2.0D, 14.0D, 16.0D, 14.0D), IBooleanFunction.ONLY_FIRST));
-		result.put(Direction.NORTH, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(2.0D, 2.0D, 0.0D, 14.0D, 14.0D, 14.0D), IBooleanFunction.ONLY_FIRST));
-		result.put(Direction.SOUTH, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(2.0D, 2.0D, 2.0D, 14.0D, 14.0D, 16.0D), IBooleanFunction.ONLY_FIRST));
-		result.put(Direction.WEST, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(0.0D, 2.0D, 2.0D, 14.0D, 14.0D, 14.0D), IBooleanFunction.ONLY_FIRST));
-		result.put(Direction.EAST, VoxelShapes.combineAndSimplify(OUT_SHAPE, Block.makeCuboidShape(2.0D, 2.0D, 2.0D, 16.0D, 14.0D, 14.0D), IBooleanFunction.ONLY_FIRST));
-		return result;
-	}
+	@SuppressWarnings("UnstableApiUsage")
```
ChloeDawn, line 37: To clarify, the unstable API here is `Maps.immutableEnumMap`, but usage warnings can be ignored, as we are within the lifecycle of the game and the Guava API will only ever change if Mojang bumps the dependency version in an update, so the usage is only as unstable as depending on game code itself.

vectorwing:
Oh my, that looks awesome! Seems like we can really do wizardry with VoxelShapes.

Though I feel the canvas stripes inside the handles look a little noisy; would probably look better using just the clean bamboo sides of the textures. This should be easy to nudge, depending on how the side models are made.

Are you still going to revisit the model (you mentioned that on Discord i think), or would you prefer this merge to be reviewed as-is? I can help, if needed!

ChloeDawn:
I will revisit the model and fix up the handle texturing

ChloeDawn:
Should be ready for merge, apologies for the delays

vectorwing (approved):
No need to apologize! These changes look amazing! :D

PR #39, opened 2020-07-31
CHS Translation

WuzgXY-GitHub:
`Hunt and Gather` does not have an official translation.
`Bonfire Lit` is from _Dark Soul_ iirc. Considering that its Chinese translation is not from the game itself, I decided not to translate it.
`Fire Up the Grill` is a little bit difficult to translate because there is no similar phrase of it in Chinese as far as I know, so I will leave it here to see if there is someone who is able to translate it without losing its own meaning.


WuzgXY-GitHub:
`farmersdelight.container.cooking_pot.served_on`
Not sure about this one, even after asking others, so it might need a double-check.

vectorwing:
First of all, thank you so, so much for this translation! Having people take the time to translate my mod is a big honor! 😄 

I'm okay with you or other translators taking liberties with difficult sentences, as I also ran into some weird decisions for the pt_br translation, even as a native speaker.
- `Hunt and Gather` is mostly a reference to ancient survival and primitive tools, as the Flint Knife is characterized as one. Feel free to fill in a fitting reference to the "stone age" or survival, if you'd like.
- `Bonfire Lit` I left untranslated in pt_br too, mostly because I never played the game in Portuguese to know the exact sentence. I'm cool with the reference staying or being overriden either way.
- `Fire Up the Grill` isn't a reference to anything; it's more of something you'd say at the start of a family barbecue. Perhaps there's a similar sentence to add in its place?

Would you prefer the current work to be reviewed as is now (and perhaps incremented later), or leave it up for more translators to take a look?

WuzgXY-GitHub:
- I thought `Hunt and Gather` was a restaurant, and I will change it soon.
- `Bonfire Lit` is shown as `BONFIRE LIT` even in Chinese so I think I'd better not to translate this.
- I know `Fire Up the Grill` has no reference, I just meant that in China we don't have a locution or so when we start a family barbecue, so translating it might make it lose its own meaning. But I will try to make it better.

vectorwing (approved):

PR #41, opened 2020-07-31
German localization

SebastianD334:
Added German translations for everything in this build.


vectorwing (approved):
Oh, very nice! You even grabbed the newer things too.
Everything looks good! Thank you so much for the translation! 😄

PR #100, opened 2020-12-13
Make saplings grow on rich soil

ConductiveFoam:
The code that tries growing a tree checks if the block the sapling is placed on is either tagged with forge:dirt or is farmland (c.f. net.minecraft.world.gen.feature.TreeFeature#place), so the only thing that's needed is to tag rich soil appropriately


vectorwing:
Brilliant! Thank you for solving it 😄👍

PR #101, opened 2020-12-13
Fix gradlew permissions so it can actually be run in UN*X

ConductiveFoam:
Quick fix to make setting up faster for future UN*X using devs.


vectorwing:
Sorry, only saw this now. Thank you for setting it!

PR #114, opened 2020-12-27
Fix Organic Compost lighting check.

Jozufozu:
Very tiny fix. The position of the compost was used instead of the neighbors position, so it wasn't correctly finding the greatest nearby light. It also no longer considers block light, as that seems closer to the [documented behavior](https://github.com/vectorwing/FarmersDelight/wiki/Rich-Soil#crafting).


vectorwing:
Hello! Sorry for not checking this earlier, stuff got in the way. 😅 
Thank you for spotting and fixing the issue. I'll do some simulations with your changes in regards to sky light, as I wasn't sure if sky light accounted for darkness from blocks above and around.

vectorwing:
All good! 👍

Issue #117, opened 2021-01-04
TeamAbnormals mods: Compatibility Issue


COLDW0LF:
Hi there,

Love the mod!
Just wanted to let you know that the Team Abnormals mods have changed their naming system,
which has broken your compatibility; "driftwood_log_stripped" is now "stripped_driftwood_log", etc.

I hope this is helpful for you. I've been tweaking recipes and solving error logs for a custom pack using OpenLoader, 
so if you want me to upload the changes after I've done them I'd be more than happy to.

vectorwing:
Hello! Thanks for spotting it; I haven't messed with integration in a while, so this would've totally skipped me. 😄

Moments after you posted the issue, @Partonetrain posted a PR fixing the IDs, and I got it merged. If you have other recipe integrations besides those, I'd be happy to take a look!

COLDW0LF:
If I come up with any more, I'll let you know! ;)

vectorwing:
I'll close this issue for now, as I'm about to review integrations for stuff I might have missed. Thanks for reminding!

PR #118, opened 2021-01-04
Add some wood Cutting recipes for integration

Partonetrain:
Morado (Atmospheric, was missing)
Maple (Autumnity)
Netherwood (Silentgear)
Driftwood (Upgrade Aquatic, had wrong ID)


vectorwing:
Ooooh, very nice! There was an issue posted minutes ago (#117) regarding UA's block ID changes, in fact, so thank you for fixing it too! 👍

PR #120, opened 2021-01-08
Add ToolType to Cutting Board recipes

ConductiveFoam:
Cutting recipes can now have a `"type"` instead of `"tag"` in their tool specification.

KnifeItem now also has a `public static ToolType KNIFE_TOOL`.


vectorwing:
Oh my stars, you are a saint! 🙏

Excellent work overall! I'm still reading along the changes, but I did clone your repo to generate a build for testing, and it worked flawlessly so far. The `type` argument recognizes pretty much every modded axe and pickaxe perfectly, save for some exceptional cases (e.g. Tetra), which is fine.

Seriously, thank you so much for this contribution! I'll post some comments before approval, but you're going into the credits for sure. 😄

src/main/java/vectorwing/farmersdelight/items/KnifeItem.java
```diff
@@ -19,19 +20,30 @@
...
+import vectorwing.farmersdelight.FarmersDelight;
 import vectorwing.farmersdelight.registry.ModBlocks;
 import vectorwing.farmersdelight.utils.tags.ModTags;
 
+import javax.annotation.Nonnull;
 import java.util.Set;
 
 public class KnifeItem extends ToolItem
 {
+	public static final ToolType KNIFE_TOOL = ToolType.get(FarmersDelight.MODID + "_knife");
```
vectorwing, line 33: Although this works, I think I prefer treating knives with a tag.

This ToolType would exclude knives from other mods (e.g. Aquaculture, Mystical World, Druidcraft etc.) from interacting with these recipes, since ToolType integration has to be done in code, and this one is requiring the FD namespace, which most mods wouldn't be using. Tags can be customized more readily in this case, since it's an exceptional, non-vanilla kind of tool.

Could you revert it to use tags? Thanks!

vectorwing (changes requested):
Everything looks good so far! Just needs to revert knives to using a tag, as per the comment. =)

src/main/java/vectorwing/farmersdelight/items/KnifeItem.java
ConductiveFoam: Changed that, the recipes now use tags again.

vectorwing (approved):
Excellent, thank you! ^^

PR #122, opened 2021-01-10
added tetra support for knives

mickelus:
Put together a base implementation for tetra knife support, figured this would be faster/easier than you having to figure it out yourselves without docs 😅 

Notes:
* knives from farmers delight can be placed in a tetra workbench and modified (which turns them into tetra items)
* knife blades can be crafted in the tetra workbench
* knives should have the same damage and speed values as regular swords
* the "cut" tool type scales at 1:1 ratio based on the material (which is slightly better than tetra shortblades)
* once converted to a tetra item the knife doesn't look as nice as the farmers delight counterpart
* module textures has to be within "assets/tetra/textures/items/module/" for them to be added to the texture atlas (unless you use something else for loading the textures)
* all data files has to be in the tetra namespace for them to be picked up
* if you want to move the glyph for the knife into an existing texture you can define "textureX" and "textureY" fields for glyphs to specify offsets within the sprite sheet


vectorwing:
Oh wow! That's quite a thorough integration! 😄

This sounds really exciting. I'll have to reserve time in the future to review/test it properly, though. Just two questions:
- I assume this integration would permit knives made from any Tetra-compatible blade/handle material?
- Knives currently employ the unofficial `forge:tools/knives` tag to broaden cutting board compatibility with modded knives, and `farmersdelight:tools/knives` for game logic. Is there a different item ID for such shortblades? If not, is there other identifiers I could use? I assume the ToolType `cut` to exist in swords as well, so not sure if I should use it.

Thanks in advance!

mickelus:
Ah, I forgot to reply! That would make swords usable as well, you could check the level of the tool type but then wooden knives would not work (as they have the same cut level as basic blades). I'll add a reminder to add tag support in the next release, been wanting to do that for a while anyway!

vectorwing:
I finally got a chance to try this integration out on a local clone. Gotta say, the fact that your mod can conjure brand new tools out of nothing but .json's and .png's is incredible. I can make a knife out of anything!

The problem is, as fun as this integration is, those knives aren't usable as FD knives in most cases, as I'm still using a tag to check for such interactions (trying to avoid `instanceof`). I also understand the item `modular_sword` would be too broad to be tagged with them.

With some recent revelations from other mods and PR's (plus ToolType integrations on the Cutting Board), it got me thinking: should my Knives perhaps have a `knife` ToolType? Since it's purely based on a string, any mod doing the same would theoretically work, and it could enable Tetra-borne knives and shortblades to function as such. Perhaps both type and tags could be checked... I have to take a look at it sometime.

What are your plans for tagging integration? 😃

shattyyt:
Hello! Love your mod! May i ask why you don't want to use `instanceof`?

vectorwing:
@shattyyt `instanceof` is used to verify if something extends or uses a certain Java class, which would restrict it to just my own mod and no external ones. Tags, types and the like are better at opening room for cross-mod compatibility, depending on the case.

shattyyt:
Oh okay makes sense.

mickelus:
Took at stab at this today, unfortunately the itemstack is not passed to `Item.getTags` and `Item.isIn` so I can't pass tags based on which modules the item has :/.
Going with both the `knife` and `cut` tooltypes could be a neat way to get your knives to be usable as tools in tetra without having tetra swords being usable on the cutting board! Should I update the PR with that change?

vectorwing:
I swear I haven't forgotten about this PR... 😓 

Many features/fixed stepped in front of it lately, and I couldn't decide yet whether to integrate two ToolTypes on the knives. But after learning more, that might be the way to go indeed. Apparently that won't break anything.

Though I'm also wondering whether to maintain both ToolTypes and the current tags, as the latter allows datapack creators to declare more items as FD knives for crafting (`forge`) and game actions (`farmersdelight`).

I'll update here once that's worked on, if you'd like to update it. ˆˆ

mickelus:
No stress dude, I don't want you to feel any kind of pressure that you have to make a decision about this! Poke me when/if you want some changes or help with fixing the conflict, or if you wanna bounce some ideas!

mont-40:
Will this feature be integrated? It would be very nice!

mickelus:
I think it would be possible to do partial (and hacky!) tag support in tetra by adding a mixin in ItemStack.is(TagKey<Item>), I'll look into it for 1.20!

vectorwing:
As discussed with mickelus recently, his new addon, [tetracelium](https://www.curseforge.com/minecraft/mc-mods/tetracelium), accomplishes this integration, with additional features as well.

Since I have long since stopped supporting 1.16, we agreed to close this PR.

Issue #110, opened 2020-12-22
[1.16.4] [Upcoming] Rich Soil issue involving saplings and giant spruce trees


vectorwing:
**Note: This is an upcoming 1.16.3 and 1.16.4 release issue. It does not apply to 0.2.4b or earlier versions.**

**Description**
To fix issue #96, Rich Soil must be added to the tag `forge:dirt`. However, Giant Spruce trees also make use of this tag to determine which blocks can be replaced with Podzol. Because of this, all nearby Rich Soil will be converted to Podzol during a Giant Spruce growth, if added to this tag.

Since users are encouraged to place Organic Compost near blocks like Podzol to accelerate it, or might consider speeding up a Giant Spruce tree with Rich Soil, this issue could make players lose a lot of soil they've been working to get, so a fix is important.

The only promising solution so far is to insert a **mixin** to the `AlterGroundTreeDecorator` class, making it ignore Rich Soil when checking for suitable blocks. Given I am inexperienced with mixin, I'm making this issue as a reminder to study up on it, and fix it separately from other enhancements. **Help wanted.**

**Steps to reproduce:**
1. Add Rich Soil to the `forge:dirt` tag;
2. Place a cluster of Rich Soil on the ground;
3. Plant a 2x2 of Spruce Saplings near the cluster, and make it grow.

**Expected behavior:**
Rich Soil should survive Podzol conversion, while also being able to grow saplings into trees.

vectorwing:
Fixed by @Jozufozu on pull request #124.

PR #124, opened 2021-01-13
Big spruce trees no longer replace rich soil.

Jozufozu:
A solution for #110.
`Feature.isDirtAt` is *only* used in `AlterGroundTreeDecorator`, and it was easier to use mixins with, so it seemed like a reasonable target.

As a side note, in order to see this change in your dev environment, you have to refresh gradle and run `genIntellijRuns`/`genEclipseRuns`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,13 @@
 buildscript {
     repositories {
         maven { url = 'https://files.minecraftforge.net/maven' }
+        maven { url='https://dist.creeper.host/Sponge/maven' }
         jcenter()
         mavenCentral()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


vectorwing:
Nicely done! I'll test it later tonight. 👍 

Given its annotation nature, I'm not sure if it's possible, but can mixins be set to toggle through a config? Just in case it breaks some mod interaction, and the player needs it disabled to troubleshoot. Its fine if it can't.

Jozufozu:
The content of the mixin can be wrapped in another if statement, and that shouldn't interfere with anything.

vectorwing:
Just tested the changes, and they work flawlessly. Rich Soil was kept throughout all tests.
Thank you so much for the help! I'll merge it and consider the config later. 😄

PR #128, opened 2021-01-16
Add preliminary JEI integration for composting

ConductiveFoam:
Just to get the ball rolling.

I'm open to feedback about translation texts and especially the position of the accelerators.

When the position's decided on the ui background image needs adjusting, both in jei and the texture itself.


vectorwing:
Once again, excellent work! 😄
Looks just right. Since we discussed changes on Discord, I'll move on to merging.

PR #130, opened 2021-01-17
Polish language translation

przemykomo:
I translated the mod to polish language.
I translated `Steak and Potatoes` as `Niedzielny obiad` which means `Sunday dinner` as a reference to a typical polish sunday dinner.
Also:
```json
  "farmersdelight.tooltip.cooking_pot.many_servings": "Przechowuje %s porcji:",
```
In polish language we would sometimes say `porcje` or `porcji` so I wrote `porcji` which should be corrent most of the times.


vectorwing:
Thank you so much! ^^
I'll get it merged right away.

PR #135, opened 2021-01-18
Holding a cooking pot changes pose of both arms

przemykomo:
Just as in title, there is only a json cooking pot model left to do.


vectorwing:
Hey there, sorry for not replying earlier! I was taking "a break" for a bit and only looked at things now and then.
That's pretty cool! I'll try testing a two-handed pot model with this change later on, and let you know how it goes with other off-hand items. 👍

vectorwing:
Been a while, huh? 😄

So I tested this today, and at first it sounded like it'd work... but after I put the model in, the pot would just hang from the holding hand, and the other arm would swing away from the handles. Not your fault, just limitations with entity rendering.

image

There seem to be ways to work around it, like Carry On does it; but from their example, it seems to be extremely involved at the render level, so I'm not sure this will work out just for this change. At least for now. :(

I'll close the PR, but keep the idea logged for some other time. Thank you anyway!

Issue #142, opened 2021-01-22
Cutting board recipes cannot use a non-vanilla ingredient serializer for tools


ochotonida:
**Description**
This is something I noticed looking through your code while working on a pull request that coincidentally fixes this very issue (see #143). 
The serializer for cutting recipes applies a special case check to allow the `type` key in the `tool` ingredient to define tool types. Apparently forge already uses the same `type` key to allow custom ingredients serializers to be used.

**Steps to reproduce:**
Create a datapack which adds a cutting board recipe with uses a non-vanilla ingredient serializer for the tool ingredient, for example:
```json
"tool": {
  "type": "forge:nbt",
  "item": "minecraft:stick",
  "count": 2
}
```
The recipe fails to load, because the recipe serializer tries to get/create a tool type called `forge:nbt`, which is not a valid tool type name.

**Mod list:**
Farmer's Delight 0.3.1-1.16.3+

vectorwing:
Good catch. As stated on your PR, I'll wait before merging this fix, just to decide whether it warrants a 0.4.0 (since it's technically a breaking change), and whether to release the current minor updates as a dot release.

Pikabyte:
Is this why you can't have more than one cutting board recipe with the same ingredient, even if the recipe uses different tools? (This is a bug I've just encountered that may or may not be related to this issue. If you have more than one recipe with the same ingredient, one of the recipes cease to function.)

vectorwing:
That could be related, but I suspect it may be due to the naming schemes.

I assume you suffixed your alternate recipe? (example: `porkchop` (knife) and `porkchop_with_axe` (axe)) If even then they cancel out, it could be some mistake in parsing that I have to look into.

Pikabyte:
I suppose I'm not sure what you mean by "suffixed". Do you mean a different recipe name? If so, then yes, I have a different recipe name.

ochotonida:
Turns out this is actually a bug (unrelated to this issue) with how farmer's delight processes cutting recipes, so you're probably not doing anything wrong. 
I've created a PR that fixes the issue (#161)

vectorwing:
Woops, I merged the PR and forgot to close this issue. My bad. Thanks for the fix!

PR #143, opened 2021-01-22
Added ToolType ingredient serializer for cutting recipes

ochotonida:
I was browsing through your code a bit when I noticed that the cutting board recipe could probably be cleaned up a bit, since it applies a special case check for tool types. I then learned that it is apparently possible to add custom ingredient serializers, which removes the need for this special case behavior entirely. 

This PR removes the special case behavior for tool types in CuttingBoardRecipe and fixes issue #142 by adding a tool type ingredient & serializer. It should be noted that data packs which define cutting board recipes with a tool type will no longer work properly, as the `type` key in the `tool` ingredient is no longer used to define tool types (see #142).


vectorwing:
Ooh, nicely done! I did notice it using "type" (which I personally liked for readability's sake), but it does block the ability to use custom ingredients or NBT.

And yes, presumably this will now be a breaking change for pack makers, which would entail a 0.4.0 version. I'll get back to you later once I see what's going into the next dot release. 👍

ConductiveFoam:
First off, thanks! This is imo a great feature. I noticed one thing, however: Tools with dynamic types, like ImmersiveEngineering's Drill or Buzzsaw, don't work anymore with your Ingredient. There's a simple fix though, override the Ingredient's test:
```
@Override
public boolean test(ItemStack stack) {
    return stack.getToolTypes().contains(this.toolType) || super.test(stack);
}
```

ochotonida:
I didn't notice tool types can be dynamic when I implemented this, good catch!

eutro (approved):
Everything looks good here to me.

vectorwing:
Alright, everything seems good here too! Next version will be a 0.4 update, focusing on these changes, and maybe some new things.

PR #146, opened 2021-01-23
Make pantries retain name when broken

bageldotjpg:
Small change that uses the `"minecraft:copy_name"` function within the loot table to make pantries retain their name when broken, if named via an anvil. Basically just works how the barrel loot table works.


vectorwing:
I could've sworn that was working. Maybe I spent so long ensuring the pot had this ability that I assumed the pantries worked too. lol

Thanks for fixing it!

PR #150, opened 2021-01-24
Added cooked_bacon & raw_bacon item tags

ochotonida:
Title.


vectorwing:
Hmm, good call! Bacon _is_ a common addition, so it does warrant its own tag separate from raw and cooked pork.
Approved! 👍

PR #157, opened 2021-01-28
Update ru_ru.json

SeregaCthtuf:
Update Horse Feed tooltip to state tamed mounts
Update Rice Bag name ("Мешочек" -> "Мешок", bag size difference)
Update boosters description


vectorwing:
Nice, thank you! ^^

PR #161, opened 2021-01-31
Fix cutting boards unable to process multiple cutting recipes with the same input

ochotonida:
This PR fixes a bug in `CuttingBoardTileEntity` to allow them to process multiple recipes with the same input ingredient even if they have different tools. It replaces a `getRecipe` call (which only finds the first recipe matching the input) with a `getRecipes` call, which is then filtered based on the tool.


vectorwing:
Just ran these changes on a local repo, and it fixes the issue perfectly! Thank you for fixing it! 😄

PR #164, opened 2021-02-01
Update ru_ru.json

SeregaCthtuf:
Update "wolf" to "tamed wolf"


vectorwing:
Sorry for pushing that change so late in the release. x)
Thanks for the update!

PR #167, opened 2021-02-04
Remove built-in Abnormals compat

bageldotjpg:
Since we have an upcoming compat mod with Pantries, Cutting, and Cooking recipes, this PR removes the current existing ones, because there are a decent amount of recipes missing, and so that we won't have to make a PR or wait on a release whenever we add new content that needs compatibility!


vectorwing:
This makes sense. You reminded me that I wanted to update Create for its compat, but having control of your own IDs sounds better. I'm also eager to see what you folks come up with! 😄

PR #168, opened 2021-02-04
Make SoupItems configurably stackable to 16

bageldotjpg:
This PR makes Vanilla SoupItems stackable to 16, using a mixin.
The mixin reuses a lot of the code from `ConsumableItem` to work.

The change has a config setting in `Configuration` and it decides what can stack based on an item tag called `stackable_soup_items`.
I have tested the build itself as well as using the tag in both a development environment and in a modpack and it runs fine on both.

The question is, should the tag be a whitelist or a blacklist? It is currently a whitelist but a blacklist could be more beneficial.
I can't see many issues arising from making all SoupItems stackable *but* there could be a risk.
Making it affect all by default is probably something people would want more, but it also does have a very slight chance of risk.

That being said, it could affect all by default, and as issues are rarely found (if at all), they could be added to the blacklist. 
Having it natively support all SoupItems with a very slight risk of the issue is probably more beneficial to the user than only supporting vanilla ones with no chance issue, but making the user request or add compatibility on their own. 

Another slight thing to mention is if this should be disabled by default. I don't personally think it should, but that's up to you.

If you have any nitpicks or things that should be changed, improved, or removed, please let me know.


vectorwing:
Nicely done!

Yeah, this particular feature I'll have to think over a bit. Having FD introduce a "core change" such as this right away can be a small can of worms without care, given the mod didn't do such changes before.

In particular, I'm partial to affecting an allow-list first. Previously, my Comfort compat affected every single SoupItem, which directly made other modded SoupItems grant the effect when it made no sense, and the config was all-or-nothing; hence the current `comfort_foods` tag (which I'm also not so sure about, since it lacks the finesse of effect duration). For a change such as this, I'd rather have it default to vanilla only, but be open to configuration for other cases if the pack maker wants it.

I wonder, would a config to _toggle_ it from an allow-list to a deny-list work, or would it be confusing, given the tag cannot be renamed on the fly?

bageldotjpg:
That would definitely work, perhaps there could be two different tags but it switches which one is used based on the config?

bageldotjpg:
Alright, I've pushed the ability to switch it from allow-list to deny-list. If there are any nitpicks or slight changes to things that need to be done let me know!

src/main/java/vectorwing/farmersdelight/setup/Configuration.java
```diff
@@ -56,6 +59,12 @@
 		COMFORT_FOOD_TAG_EFFECT = COMMON_BUILDER.comment("Should items inside the tag 'farmersdelight:comfort_foods' grant 5 minutes of Comfort when eaten? (defaults to vanilla SoupItems)").define("comfortFoodTagEffect", true);
 		RABBIT_STEW_JUMP_BOOST = COMMON_BUILDER.comment("Should Rabbit Stew grant users the jumping prowess of a rabbit when eaten?").define("rabbitStewJumpBoost", true);
 		DISPENSER_TOOLS_CUTTING_BOARD = COMMON_BUILDER.comment("Should most vanilla tools register a dispenser behavior when facing a Cutting Board?").define("dispenserUsesToolsOnCuttingBoard", true);
+
+		COMMON_BUILDER.comment("Stack size overrides").push(CATEGORY_OVERRIDES_STACK_SIZE);
+		STACKABLE_SOUP_ITEMS = COMMON_BUILDER.comment("Should soup items become stackable to 16 like Farmer's Delight's soups & stews?", "By default, this override only affects items inside the `stackable_soup_items` tag, which starts with vanilla bowl foods.", "The setting below this reverses the behavior, overriding all registered SoupItem's instead. This includes SoupItems from other mods, so be careful!").define("stackableSoupItems", true);
```
vectorwing: The third comment line in this config might be better suited for the one below (OVERRIDE_ALL_SOUP_ITEMS), as it describes it directly.

vectorwing (commented):
I tested this locally, and it works flawlessly. Yet another excellent work!
Just a quick comment about the config comments and we're good to merge. I can change it myself if you prefer. =)

src/main/java/vectorwing/farmersdelight/setup/Configuration.java
bageldotjpg: Good call 👍

PR #176, opened 2021-02-09
Add smoked ham as a possible drop to pigs & hoglins

### LINKED ISSUE #172: [1.16.5] Pigs on fire, when killed with a knife, drops regular ham instead of smoked ham
Pigs on fire drops regular ham instead of smoked ham. This also goes for fire aspect.I don't really know what else to put here lol

ConductiveFoam:
Fixes #172


ConductiveFoam:
Not happy with how I did it, will open a better one later

PR #177, opened 2021-02-09
Add smoked ham as a possible drop to pigs & hoglins

### LINKED ISSUE #172: [1.16.5] Pigs on fire, when killed with a knife, drops regular ham instead of smoked ham

ConductiveFoam:
Properly now

Fixes #172

- Add "farmersdelight:smoker_cook" loot function that applies a SmokingRecipe to the drop, analogously to Vanilla's Smelt function
- Apply "farmersdelight:smoker_cook" to pig & hoglin drops when they're on fire


vectorwing:
Just tested, looks good! Thank you so much! 😄

PR #178, opened 2021-02-11
Override IPlantable.getPlantType on TomatoesBlock

### LINKED ISSUE #174: Can't plant tomato seeds on Botania farmland
I'm honestly not sure if this is a mistake on Botania AIOT's part or if it's a problem with Farmer's Delight but I'm unable to plant tomato seeds from FD on farmland tilled by any Botania AIOT hoe. Tested with a vanilla hoe and a Mekanism hoe, both of which work fine. Every other seed in the mod is planted on Botania AIOT farmland without issue, it's just the tomato seeds.
Any chance this can be looked into? 
Running 1.16.5 version in the Valhelsia 3 modpack. I'm not sure what other info to provide, sorry :s

ConductiveFoam:
Fixes some modded farmland interactions (notably AIOT Botania's)

Fixes #174 174


vectorwing:
Hmm, well noted. Tomatoes would indeed fail on plant type, since they're a custom implementation. Thank you for fixing!

PR #220, opened 2021-03-08
Added support for Just enought effects

DaLumma:
Support for [Just enough effects](https://www.curseforge.com/minecraft/mc-mods/just-enough-effects) could be useful for modpack players or players in general that don't know what these effects do.
JEE


vectorwing:
Good idea! However, I might write these tooltips myself, to match the general writing style. Thank you anyways for the proposal! 😄

PR #239, opened 2021-03-25
Reimplement tags for Stackable SoupItems

bageldotjpg:
By using the power of Forge's IOptionalTags, Stackable SoupItems can be defined in tags now and not break on the server or with JEI!


bageldotjpg:
Did a bit more testing on client, server, and with JEI and it seems to work without issue

vectorwing:
Pretty nice. Though I'm thinking a bit on this, and whether keeping it as a config would be easier for the end user.

This week I had to assist a player that wanted to integrate Doggy Talents dogs as a consumer of Dog Food, and realized just how many hoops they would have to jump through to have it working, as opposed to just typing the ID in a config array.

Granted, this could be advantageous in _this_ case, as we're dealing with larger quantities of items, but it's a design topic to consider. What do you think?

bageldotjpg:
I can definitely see where you are coming from, tags / data-packs do take a bit to learn.

I do think a tag is very advantageous because of the ability for mod devs to add stuff on their side. Making the user have to do all compatibility in their config if they want other mod's items to stack is not very intuitive, but with a tag, the mod devs can add support on their end or you can use the optional tag support on your end as well. 

Also as Mojang adds more and more tags & data-driven content, it might be good to encourage users to learn stuff like tags early on! (and if people still don't want to do that, things like CraftTweaker or websites to make stuff like this exist to make that a bit easier)

vectorwing:
Sorry for leaving this PR open for so long; I wasn't sure at all whether to go for it or not.

After some discussions I had, I decided to not merge this, at least not for now. Tags are finicky with the way they become bound, and there are situations where other mods might access an item stack limit before they are. This has the potential to cause problems down the line.

Thanks for taking the time to reimplement it, though! I might look at it again in the future.

PR #240, opened 2021-03-25
Convert Cabbage Roll ingredients to a tag

bageldotjpg:
This PR adds a tag for `cabbage_roll_ingredients` so that it is easier for people to add more ingredients to cabbage rolls seeing as they are such a versatile recipe!


src/main/java/vectorwing/farmersdelight/utils/tags/ForgeTags.java
```diff
@@ -40,6 +40,8 @@
 	public static final ITag.INamedTag<Item> MILK_BUCKET = forgeItemTag("milk/milk");
 	public static final ITag.INamedTag<Item> MILK_BOTTLE = forgeItemTag("milk/milk_bottle");
 
+	public static final ITag.INamedTag<Item> MUSHROOMS = forgeItemTag("mushrooms");
```
vectorwing: Just a heads-up: `forge:mushrooms` is already an official Forge Tag, and it's working on the current version; we likely don't need to fill it. 😄

vectorwing (changes requested):

src/main/java/vectorwing/farmersdelight/utils/tags/ForgeTags.java
bageldotjpg: I did try to use it, but runData failed because I suppose it wasn't 'built' yet or something

vectorwing: Hmm, that's weird... Usually I refer to official Forge tags when datagenning some recipes (such as Shears-based cutting, or the current Cabbage Rolls recipe). Perhaps it bugs out when using them to fill new tags? Not sure if I ever had to do this in specific yet.

bageldotjpg: After doing some research I've discovered this is an issue solved by Forge version 35.1.3 but since FD is not on that internally it does not work without crashing

vectorwing: Good catch! I haven't updated it for a while, mostly because I wanted to continue supporting 1.16.3. Do you know if Forge 35 and 36 is backwards-compatible on build? I'm not very savvy on that front.

bageldotjpg: Sorry, forgot to respond. Not sure. I assume this would work fine since there shouldn't be anything related to the data generators when the mod is built but you never know. I'll take a look later

bageldotjpg: Ok, after testing it out: the mod builds fine if you aren't using 35.1.3, but to runData you'll need to have the workspace in 35.1.3. So you don't have to force a new version. (although, I don't think forcing a later version would hurt since most people are on *at least* 1.16.4, and should really be on 1.16.5.

vectorwing (approved):

vectorwing:
I ended up moving this version entirely to 1.16.5, using Forge 36.1.0 as a base. Figured it was due time, especially since I needed important fixes from Forge to use loot modifers.

I'll merge this on 1.16.3 and move it to 1.16.5 later. 👍

PR #247, opened 2021-04-01
Create zh_tw.json

ghost:


vectorwing:
Hi! Your lang file is currently on the root of the project; it needs to be in the correct lang folder in order to work in-game.
Could you please move it to the folder path below? We already have a zh_tw, so you can modify it. 😄 

https://github.com/vectorwing/FarmersDelight/tree/1.16.3/src/main/resources/assets/farmersdelight/lang

vectorwing (changes requested):
@Pancakes0228 Could you please migrate your updates to the existing `zh_tw.json` inside the `lang` folder? I posted a link to the path in my previous comment.

vectorwing:
I added your changes manually to zh_tw.json in commit b1956eb. Thank you for the update!

PR #262, opened 2021-05-02
Add Configured and Catalogue Support

Dradon7:
Set the icon to the Cooking Pot when catalogue is installed, instead of the grass block, and set the background of Configured's config screen to bricks, since that's the main aesthetic of FD.

As far as I can tell, its literally a 2 line addition in order to add support for this. Lemme know if I missed something, or if you guys have better ideas for what the icons should be.


vectorwing:
I didn't know these two mods existed, but they seem to be quite convenient! Just tested here, and the config page looks quite nice. Thanks for the suggestion, I'll merge it now!

PR #263, opened 2021-05-02
Added indonesian translate (id_id.json)

arikhn:
Indonesian Translate


vectorwing:
Thank you so much for the translation! 😄

PR #283, opened 2021-06-14
Update ru_ru.json

SeregaCthtuf:
Added translation for container.cooking_pot.not_heated, container.cooking_pot.heated, tooltip.hot_cocoa and tooltip.milk_bottle


vectorwing:
Thank you for the swift update! ^^

PR #298, opened 2021-07-11
Create Compat

Dradon7:
I added compat with Create's mechanical saw
Important things I did:
 - Added cutting recipes with create's mechanical saw for meats, pies, cakes, mushroom colonies, cabbages, pumpkins, and melons
 - All meat gives one extra slice when cut so that this is more efficient than hand cutting (maybe nerf this?)
 - Cooked chicken/fish/ham have an addition piece of meat instead of bones (for a total of 4 pieces)
 - Cutting cabbages also gives seeds because extra products are easy enough to get rid of using brass tunnels (I definitely need to nerf this because as I'm writing up this PR, I realized that would make infinite seeds)

### DIFF
--- a/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/apple_pie.json
+++ b/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/apple_pie.json
@@ -1,0 +1,17 @@
+{
+  "type": "farmersdelight:cutting",
+  "ingredients": [
+    {
+      "item": "farmersdelight:apple_pie"
+    }
+  ],
+  "tool": {
+    "tag": "forge:tools/knives"
+  },
+  "result": [
+    {
+      "item": "farmersdelight:apple_pie_slice",
+      "count": 4
+    }
+  ]
+}
--- a/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/apple_pie.json
+++ b/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/apple_pie.json
@@ -1,17 +1,15 @@
 {
-  "type": "farmersdelight:cutting",
+  "type": "create:cutting",
   "ingredients": [
     {
       "item": "farmersdelight:apple_pie"
     }
   ],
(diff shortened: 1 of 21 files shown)


Dradon7:
Fixed the cabbage thing

Dradon7:
Oh, I also added a folder for create's newest recipe type, create:sequenced_assembly, but couldn't think of anything to add. Also, those recipes are very very complicated and I don't feel like learning about them this late at night.

DaLumma:
I have a short question: why?

This compat seems a bit unnecessary and unbalanced to me.

all of the things are already possible to be automated and this PR would just make it a small bit easier to automate it while giving you more meat.

vectorwing:
Thanks for proposing new integrations. 👍

I'm not sure I like the idea of the Mechanical Saw granting _even more_ meat cuts than manual means. Mechanical processes are impartial, non-artisanal; they shouldn't be necessarily more efficient than hand-cutting meat. Also, this seems like too many resources for these few common items to give.

And yes, all of those can already be automated through the use of Deployers on the Cutting Board. Although the Deployer will end up adding/removing the tool if it moves too quickly, I consider that to be part of the automation puzzle, since the Cutting Board provides a very simple comparator signal to be consulted. You would simply need to stop it when out of items, or tune its speed to be just right. 😉

Dradon7:
I could just delete the meat recipes entirely. The pumpkin/melon/pie/cake splitting recipes were the main reason I made this PR, as their cut forms seemed useful, but hard to automate. The meat stuff was added afterwards.

However, I forgot about deployers. Like Vector said, this can already be automated through the use of Deployers on the Cutting Board, making this interesting but unnecessary.

Feel free to close this

vectorwing:
After some thought, I decided not to add these. Mostly due to the Mechanical Saw being more oriented towards processing harder materials (although if we had deli, it could stand-in for a deli slicer... heh).

Sorry about it, mate! But thank you for taking the time to prepare a PR with your proposal! ^^

PR #307, opened 2021-08-11
Updated pl_pl

jogurciQ:
cleaned up the file to look like en_us.json, added missing lines, fixed some literal translations that felt out of place


vectorwing:
Looks good! Thank you for updating this translation!

PR #330, opened 2021-10-13
Updated German Translation to current state of 0.5

DaLumma:


vectorwing:
Nice one! Thanks for the quick lang update! 😄

PR #344, opened 2021-11-06
Update CHS translation to 0.5.1

WuzgXY-GitHub:


vectorwing:
Thanks for the update!

As for the changes you suggested for the en_us lang - "other" was intentional in this case. The Skillet can be operated from either hand: your ingredient can be on your main-hand, so long as you're not currently able to consume it.

src/main/resources/assets/farmersdelight/lang/en_us.json
```diff
@@ -249,7 +249,7 @@
   "farmersdelight.block.cutting_board.invalid_item": "This doesn't seem cuttable...",
   "farmersdelight.block.cutting_board.invalid_tool": "Maybe with a different tool...",
   "farmersdelight.block.skillet.invalid_item": "This can't be cooked...",
-  "farmersdelight.item.skillet.how_to_cook": "Use your other hand to hold ingredients!",
```
vectorwing: The use of "other" was intended in this case, as the Skillet can be operated from either hand. Can you please revert the change?

vectorwing (changes requested):

WuzgXY-GitHub:
Oh yeah, I tested it in creative mode, and foods are always edible in this mode. This is completely due to my carelessness, sorry for that.
~~I'm currently far away from my PC, and I forgot the password (which is stored in my PC browser) to log into my GitHub account, could you please edit the file by yourself? I think it is maintainer-editable. Sorry again, for inconvenience I brought.~~
I have reverted the change and updated the Chinese language file.

vectorwing:
Sorry for the really long delay! I got wrapped up into a break and forgot to reply to you. :(

I was going to say it could be done on my end, but thanks for taking the time to revert it. I'll go ahead with the merge. Thanks for the update once again! 👍

vectorwing (approved):

PR #345, opened 2021-11-10
Update Catalan language again

VerdaPegasus:
Update catalan language for 1.17.1, more accurate translation


vectorwing:
Thanks for the update!

PR #354, opened 2021-12-07
Add data-driven pantry registration

Snownee:
This PR adds a feature letting modders add their own pantries without the need of adding FarmersDelight as their dependency.

Modders just need to provide resources, recipes and loot tables, and simply add two lines in their `mods.toml`:

```toml
[modproperties.modid]
    farmersdelightPantryCompat=[ "modid:custom_pantry" ]
```


Snownee:
Here is an example that makes use of this PR (requires Kiwi mod):

[libs.zip](https://github.com/vectorwing/FarmersDelight/files/7675737/libs.zip)

PR #360, opened 2021-12-12
[Tiny] Fix CHS language file space problem which is caused by my auto 16-color translator

WuzgXY-GitHub:
Sorry for that :(


vectorwing:
Nice catch 👍

PR #382, opened 2022-01-15
Add more Create-integration recipes

hedgehog1029:
Add more recipes using Create machines:

- Cabbage leaves are now made by running a cabbage through a mechanical saw instead of the mixer
  - I'm not sure which is funnier, but the idea of running a cabbage across a wood saw is tickling me.
- Add sawing recipes for all the meat-doubling recipes
  - Each recipe follows the normal cutting board recipe, but with a 20% chance to produce an additional meat cut.
  - Minced beef is made in the mill instead, obviously
- Add filling recipe for milk bottles (250mB of milk, as implied by other recipes)
- Add sawing recipe for pumpkin slices
- Add several different recipes for making raw pasta:
  - Mixing 4 wheat with liquid water = 2 pasta
  - Mixing 1 dough and 1 egg = 1 pasta
  - Cutting 1 dough on the saw = 1 pasta (you know, like a... pasta roller? kind of?)

Happy to take feedback and adjust any of these!

### DIFF
--- a/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/bacon_from_cutting.json
+++ b/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/bacon_from_cutting.json
@@ -1,0 +1,16 @@
+{
+  "conditions": [
+    {
+      "type": "forge:mod_loaded",
+      "modid": "create"
+    }
+  ],
+  "type": "create:cutting",
+  "ingredients": [
+  	{ "item": "minecraft:porkchop" }
+  ],
+  "results": [
+  	{ "item": "farmersdelight:bacon", "count": 2 },
+	{ "item": "farmersdelight:bacon", "count": 1, "chance": 0.2 }
+  ]
+}
--- a/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/cabbage_slice_from_cutting.json
+++ b/src/main/resources/data/farmersdelight/recipes/integration/create/cutting/cabbage_slice_from_cutting.json
@@ -5,7 +5,7 @@
       "modid": "create"
     }
   ],
-  "type": "create:mixing",
+  "type": "create:cutting",
   "ingredients": [
     {
       "item": "farmersdelight:cabbage"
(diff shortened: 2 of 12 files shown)


Ch1216:
Don't think you should get extra meat cuts for automation, [create already adds a compat recipe for farmers delight's milk bottle](https://github.com/Creators-of-Create/Create/blob/a92855254c9a7b85ba28781e2e3ce7169549cbf7/src/generated/resources/data/create/recipes/filling/compat/farmersdelight/milk_bottle.json)s, and don't really agree with raw pasta recipes (don't think they should be added as they don't make sense(maybe the saw one makes sense)), also what about making a mixing recipe to make 3 wheat and a water bucket make wheat dough

hedgehog1029:
That recipe was added to Create 28 days ago, and this PR is a lot older than that.

I'm happy to adjust if Vectorwing is actually interested in the changes but for now I'll leave it as is.

MarioSMB:
Create has Wheat Flour which is usually turned to Dough but could be used as a means of creating pasta perhaps.  
The two mods also provide their own version of dough which is a little funky for compatibility (Create's dough cannot be used in wheat dough recipes and vice versa).

I also don't think it's such a bad thing to have Farmer's Delight handle the milk bottle recipe - it's a FD item using Create's station. Probably better than Create trying to support all possible mod items with their stations...

vectorwing:
Heyo! Sorry for the long delay on my answer, other things came up.
Hmm... To be honest, I'm not sure on most of these. My thoughts below:

> Add sawing recipes for all the meat-doubling recipes [...]

The Mechanical Saw doesn't seem to me like a workstation which should be handling food in general; it's very heavy-duty and wood-oriented. You can already automate the Cutting Board with Create by using Deployers with knives, so I don't see the need for it.

> Add several different recipes for making raw pasta:
> - Mixing 4 wheat with liquid water = 2 pasta
> - Mixing 1 dough and 1 egg = 1 pasta
> - Cutting 1 dough on the saw = 1 pasta (you know, like a... pasta roller? kind of?)

Also not sure about these. Raw Pasta would need to either be finely sliced by a Knife (artisanal), or ran through a more fitting industrial pasta cutter, which Create doesn't seem to have a great candidate for. The Mechanical Mixer is more of a "make it homogenous" machine.

> Add filling recipe for milk bottles (250mB of milk, as implied by other recipes)

This one was recently added from Create's end. I have a few plans for fluid handling in FD, which might also involve rethinking milk bottling, so I'm leaving this one for the future.

Sorry for the negative feedback. 😓 I did run a few ideas in regards to Create integration already, and haven't found much more at the moment. But thank you very much for the proposals anyway!

PR #404, opened 2022-01-28
Update it_it.json (Italian lang)

SimGitHub5:


vectorwing:
Thank you for the translation! Sorry for the delay, I was on break. ^^

PR #419, opened 2022-02-15
Make wild rice loot table use `forge:shears`

FortressNebula:
Right now only the lower half uses the `forge:shears` tag, while the upper block uses `minecraft:shears`. This leads to some unexpected weirdness when using modded shears, so I changed it to use `forge:shears` for both.


vectorwing:
Whoops, that's an oversight. Thanks for spotting it!
I also noticed it's broken on 1.16.5 too, so I'll replicate it there later.

PR #428, opened 2022-02-28
Candle Cake Fix

Umpaz:
fixes how FD interacts with modded candle cakes, specifically the cutSlice event


bageldotjpg:
Personally I'd recommend naming the tag something like `farmersdelight:drops_cake_slice`. This is a bit specific to me since I add Vanilla Cakes, so `vanilla_candle_cakes` is a slightly confusing name overlap since it could mean cakes from Vanilla or made out of Vanilla.

src/main/java/vectorwing/farmersdelight/common/item/KnifeItem.java
```diff
@@ -81,7 +81,7 @@ public static void onCakeInteraction(PlayerInteractEvent.RightClickBlock event)
 			BlockState state = event.getWorld().getBlockState(pos);
 			Block block = state.getBlock();
 
-			if (BlockTags.CANDLE_CAKES.contains(block)) {
+			if (BlockTags.DROPS_CAKE_SLICE.contains(block)) {
```
vectorwing: This line is pointing to the vanilla tags class, which doesn't contain your new mod tag.
You gotta change it to `ModTags.DROPS_CAKE_SLICE`.

vectorwing (changes requested):
Sorry for the long delay; finally got time to try this one out.

Yeah, I agree that `drops_cake_slice` is a nicer name, as it becomes a template for other cake slices too. Everything here works fine, just found a small error that needs fixing. 👍

vectorwing (approved):

Issue #432, opened 2022-03-12
Enhancement to ropes: Have a way for rope to be collected from the top


jjac31415:
Hello,

I'm not sure if this would be possible, but it would be nice to be able to gather the rope back into your inventory from the top. Perhaps right click or sneak-right click with an empty hand on the top segment of rope would gather the lowest segment back into your inventory.

I have no idea how hard this would be, and even without it the rope is a very cool item :)
Thank you for this wonderful mod.

Jerry

vectorwing:
This was actually in the plans, yeah! After a lot of play, I noticed ropes weren't easy to retrieve, like a water bucket can be sometimes.

LimonBlaze opened a PR introducing something similar to what I envisioned. I'll take a look and discuss it with them first and see where it goes; if it's approved, it will likely be on a minor release.

vectorwing:
Added on commit 1538311, thanks to @LimonBlaze's PR contribution. Should be out as a feature in the next release.

PR #439, opened 2022-03-25
Basically satisfied #432: Made ropes collectable from top

RaymondBlaze:
This PR solved #432 : Ropes now can be collected from top using Shift + Right Click.

Additional information:
1. Player must have its mainhand empty to do this.
2. Collected ropes will be merged into existing rope item stack in inventory (if present) or added into the first empty slot in inventory (except for both hands, because having rope item in offhand leads to the action of placing it, which is unexpected).
3. I didn't check whether the clicked rope block is the upper end. From my POV, one can pull ropes from its middle for sure, and this function would be too useless if we add such restriction.
4. The idea of collecting ropes from bottom is unrequired by the issue, and I haven't got any reason to do so.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/RopeBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/RopeBlock.java
@@ -4,7 +4,9 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.context.BlockPlaceContext;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.Level;
@@ -21,6 +23,7 @@ import net.minecraft.world.phys.BlockHitResult;
 import net.minecraft.world.phys.shapes.CollisionContext;
 import net.minecraft.world.phys.shapes.Shapes;
 import net.minecraft.world.phys.shapes.VoxelShape;
+import vectorwing.farmersdelight.FarmersDelight;
 import vectorwing.farmersdelight.common.registry.ModBlocks;
 
 @SuppressWarnings("deprecation")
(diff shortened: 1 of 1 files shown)


vectorwing:
Hi there! Thank you very much for contributing this feature. ^^

You sorta nailed what I wanted to achieve with it: points 1 and 2 are solid, and 3 seems fair to avoid a pointless restriction. As for 4, bottom-end reeling would be a lot more useful if ropes collapsed without support (something I haven't yet managed to add), so it's fine without it.

I only have a few revisions after playing around with the code, mostly for optimization.

src/main/java/vectorwing/farmersdelight/common/block/RopeBlock.java
```diff
@@ -46,21 +49,59 @@ public BlockState getStateForPlacement(BlockPlaceContext context) {
 		return state != null ? state.setValue(TIED_TO_BELL, world.getBlockState(posAbove).getBlock() == Blocks.BELL) : null;
 	}
 
+	public int findSlotForRope(Inventory inventory) {
```
vectorwing: This function isn't necessary for this use case. You can achieve a similar effect by using `player.inventory.add(itemStack)`, as it returns a boolean for whether it succeeded or failed to add the item. This version of the method will already find the first available slot to add the item to, as well.

You can add this check alongside the `getAbilities().mayBuild` check, as both must be true together. This eliminates the `if (ropeSlot > -1)` block, and the inventory field.

vectorwing: Also, as for potentially failing the block removal after adding the item, this likely won't happen, as the player won't have a free hand once the last slot is filled with a Rope, preventing further reeling anyway.

vectorwing (changes requested):

src/main/java/vectorwing/farmersdelight/common/block/RopeBlock.java
RaymondBlaze: The reason why I didn't use `Inventory#add(itemStack)` is that this method would always place the stack in player's **first empty slot** if there's no stack to combine. In that case, it would cause slightly misfunction as the first empty slot would most likely to be the **main hand**. Placing stack in main hand would block the reeling action, and the player would have to stop and change main hand to another slot.
Considering that the chance of a player to reel the rope without holding any is actually quite small, I would like to change my original idea and follow your suggestion :)

vectorwing: That's actually a fair point, didn't think of that. It would most likely prioritize the hotbar, speeding up the blockage of an empty hand, so actually looking into the inventory first makes sense.

I'll decide later which one to go for, but thanks for updating either way. The method could probably reside in ItemUtils as well, given it may be useful for other cases.

vectorwing:
Hi again! Sorry for the delay.

I decided to come back to this PR and test it again, regarding the discussion we had about leaving a hotbar slot free when reeling ropes. Turns out that the `getInventory().add()` method takes this into account when selecting which slot the new item goes to.

If the only empty hotbar slot is the selected slot, it will instead attempt to place it in the player's inventory, only resorting to the selected slot **last**. I tested locally, and it kept my hand free up until my inventory was completely full, and only my empty hand was left. Here's the test, with the arrow marking my (once) empty hand:

image

vectorwing (approved):

vectorwing:
@LimonBlaze, the changes are great! However, since this was aimed at 1.18.1 (again, sorry for taking so long to reply), it might be a bit outdated. I can add these changes from my end and just close the PR, while crediting you as usual, if you prefer; since a lot has happened between this and the 1.2 dev branch. 😅

RaymondBlaze:
>  I can add these changes from my end and just close the PR, while crediting you as usual, if you prefer;

It's fine for me, handling a PR on an outdated branch is truly not a good idea, and I made the PR just for being helpful and improve the mod :)

vectorwing:
Alright! Added this feature on commit 1538311. Thank you very much for the contribution! 👍

PR #440, opened 2022-03-25
Fixed #433: Introducing biome specific wild crops generation configs

### LINKED ISSUE #433: [1.18.1] [Bug and/or suggestion] Wild crop dimension blacklist
**Description**
Wild crops can generate in certain modded End (and Nether, I assume) biomes, such as BYG's Nightshade Forest. I'm guessing it's because those biomes have climate tags that match Farmer's Delight's requirements?

**Expected behavior:**
I assume that crops are only meant to generate in the Overworld, but either way an option to blacklist biomes and dimensions would be helpful.

**Screenshots**

RaymondBlaze:
This PR fixed #433 by introducing 2 more configs:
1. (Boolean) Generate wild crops in vanilla biome only
2. (List<String>) Wild crops biome blacklist

Dimension specific generation config is kind of unrealizable, so I made the first one instead. It's kind of duplicated with the second one, but may be useful in certain conditions (e.g. the player added too many biome related mods and is exhausted to test and blacklisting every possible biomes).

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/Configuration.java
+++ b/src/main/java/vectorwing/farmersdelight/common/Configuration.java
@@ -1,9 +1,11 @@
 package vectorwing.farmersdelight.common;
 
 import com.google.common.collect.ImmutableList;
+import net.minecraft.resources.ResourceLocation;
 import net.minecraftforge.common.ForgeConfigSpec;
 import net.minecraftforge.fml.common.Mod;
 
+import java.util.ArrayList;
 import java.util.List;
 
 @Mod.EventBusSubscriber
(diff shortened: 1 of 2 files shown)


vectorwing:
Sorry for taking so long to analyze this PR. Other things took my focus during development. 😓

I had a while to think on this issue, and during that period, new solutions cropped up. Forge has implemented biome tags in 1.18.2, one of which is `forge:is_overworld`, which I can now use in a `PlacementModifier` or similar to prevent Nether/End generation.

As for this PR's solution, I don't want to restrict generation to vanilla-only biomes, as many mods introduce valid Overworld settings for wild crops. And yeah, as you mentioned, having to fill in a deny-list with each new biome addition isn't scalable.

I'm currently working on the aforementioned PlacementModifier solution; I'll come back to this PR after testing it. 👍

vectorwing:
The PlacementModifier strategy ended up working. See commit fae0b24 for details.
With it, Wild Crops are now bound to overworld-only, which fixes the issue's original example with BYG.

Regardless, thank you very much for taking your time to help fix/troubleshoot the issue! I appreciate it a lot. Sorry it didn't get added. ^^

PR #459, opened 2022-04-24
Update de_de.json (1.16.5)

DaLumma:
Updated the german translation for 1.16.5


src/main/resources/assets/farmersdelight/lang/de_de.json
```diff
@@ -4,14 +4,14 @@
   "block.farmersdelight.basket": "Korb",
   "block.farmersdelight.cutting_board": "Schneidbrett",
   "block.farmersdelight.skillet": "Bratpfanne",
-  "block.farmersdelight.oak_pantry": "Eichenholzküchenschrank",
```
vectorwing: It seems you used the new 1.17+ keys in this lang. 1.16 still uses the old keys, such as `pantry` and `rice_crop`.
All diff lines that change keys are affected, so it should be easy to rollback.

vectorwing (changes requested):
Thanks for the update! I commented on an issue with the lang keys; we can merge once resolved.

src/main/resources/assets/farmersdelight/lang/de_de.json
DaLumma: they should now be back to the old ones didn't notice that they stayed untouched in 1.16

vectorwing (approved):
Nicely done! 👍

PR #460, opened 2022-04-24
Update de_de.json (1.18.2)

DaLumma:
Updated the german translation for 1.18.2


vectorwing:
Thank you very much! Everything looks good here. 👍

PR #494, opened 2022-06-26
Update languaje ca_es 

VerdaPegasus:
translated new entries and deleted old ones,


vectorwing:
Thank you very much! 😄

PR #495, opened 2022-06-26
Add Straw Tag for Recipe

0999312:
Add tag "forge:straw" for Straw that can make other mods easiler to compat or something.


vectorwing:
Hmm! Although this might be a good idea, I admit I'm not familiar with other mods which add straw. I'd probably have to look into some use cases first, as well as consider balance implications with mods that do add it.

Also, make sure you `runData` when doing changes to datagen code, as we upload the generated data to the repo as well.

0999312:
Err...... OK runData soon

0999312:
OK, long time passed, there we go. :grin:

vectorwing:
Hi! Sorry for the long delay.

After thinking a bit, I decided not to include this tag. Some of those recipes use Straw in very generic shapes, which might overlap with another mod's own usages of straw. Sorry, but thank you for the PR either way!

PR #500, opened 2022-07-04
Create cabinet item tag for crafting

brnbrd:
This PR introduces a common tag for all Cabinet BlockItems
"farmersdelight:cabinets", with wooden ones tagged "farmersdelight:cabinets/wooden" (all of the ones in the base mod).
 This mirrors Forge's chest tag implementation of "forge:chests" and "forge:chests/wooden".
I suppose the "farmersdelight" namespace could be swapped with "forge" to make it more general, if that is preferred.
This tag is helpful for modders who wish to add crafting recipes accepting any cabinet item, and works to establish a general tag that compat mods (such as Every Compat) can dynamically add items to.


vectorwing:
Hmm... yeah, this could work. I feel it's fine to just leave it in the mod namespace, as FD's idea of a "cabinet" isn't universal. If it helps with crafting or organizing/searching, then it's a good tag to have. 👍

PR #514, opened 2022-07-30
Add support for `forge:dough` and `forge:dough/wheat`

ChloeDawn:
This allows Farmer's Delight dough to be used in Create recipes, and Create dough to be used in Farmer's Delight recipes.
The ingredient for the recipes has been set to the wider scope of `forge:dough` as this is also what Create uses.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/tag/ForgeTags.java
+++ b/src/main/java/vectorwing/farmersdelight/common/tag/ForgeTags.java
@@ -37,6 +37,9 @@ public class ForgeTags
 	public static final TagKey<Item> CROPS_RICE = forgeItemTag("crops/rice");
 	public static final TagKey<Item> CROPS_TOMATO = forgeItemTag("crops/tomato");
 
+	public static final TagKey<Item> DOUGH = forgeItemTag("dough");
+	public static final TagKey<Item> DOUGH_WHEAT = forgeItemTag("dough/wheat");
+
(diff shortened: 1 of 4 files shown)


vectorwing:
Thanks for accidentally reminding me to add this! 😆

I was discussing this tag with the Create team before 0.5, but sorta forgot to implement it from my end too. Yeah, this is a good addition, as it expands on dough items viable for cutting Raw Pasta.

Could you `runData` from your end, so I can add the .jsons to the repo already?

vectorwing:
Actually, I can `runData` from my end, that's okay.
Thanks for the contribution!

ChloeDawn:
Sorry for the late response! I was having trouble running the data generators, there appears to be a bug with the cutting recipe generator and the ID generation does not support tags (it resolves to a barrier) and creates `barrier.json`.

vectorwing:
Ohhh. I think I ran into this in the past, actually... you might have reminded me again of a reason why I left it for later.

`CuttingBoardRecipeBuilder` tries to pull the ingredient's name for the filename (given it has multiple outputs), which until now is always a single non-tag item. But it's unable to grab a sample from the tag. 😔

I'll try to fix the builders and improve the saving method for the future. Thank you again for the help! 👍

vectorwing:
Yeah, it seems like tags are not populated during this stage of datagen. Since it's empty, the game tacks on a barrier item to it, and that's all it can find when getting the ingredient's first item.

Gonna have to figure out a different auto-name, or throw an exception. 😓

PR #524, opened 2022-08-14
Add a Recipe Book to the Cooking Pot's interface

vectorwing:
- The Cooking Pot now has its own **Recipe Book**!
  - Opening the GUI will reveal the familiar green book button. Press it to see your current cooking recipes;
  - Recipes are unlocked in the same fashion as vanilla: by having one of its ingredients, or cooking the recipe from memory once;
  - Added optional `"recipe_book_tab"` field for cooking recipes, which lets the recipe specify which tab it goes into:
    - Valid options are `meals, drinks, misc`;
    - If not provided, the recipe's tab will be determined in-code;


src/main/java/vectorwing/farmersdelight/common/Configuration.java
```diff
@@ -75,6 +78,11 @@
 				.defineList("canvasSignDarkBackgroundList", ImmutableList.of("gray", "purple", "blue", "brown", "green", "red", "black"), obj -> true);
 		COMMON_BUILDER.pop();
 
+		COMMON_BUILDER.comment("Recipe book").push(CATEGORY_RECIPE_BOOK);
+		ENABLE_RECIPE_BOOK_COOKING_POT = COMMON_BUILDER.comment("Should the Cooking Pot have a Recipe Book available on its interface?")
+				.define("enableRecipeBookCookingPot", true);
+		COMMON_BUILDER.pop();
```
vectorwing, line 84: TODO: Study a feature for automatically unlocking all cooking recipes, if a config is enabled (would be disabled by default).

vectorwing: Marked for later consideration.

PR #534, opened 2022-08-31
[Compat / 1.2] Better Combat compat for skillet.

murao-kun:
This will make Skillet work with Better Combat mod, turning skillet into claymore type weapon. (Since it is heavy and slow)


murao-kun:
Changed preset to mace. I guess that fits it a lot better.

vectorwing:
Saw the video you sent over on Discord, and heck yeah, mace is perfect for the Skillet! It even makes a clunking sound. ❤️
Thank you for the integration!

PR #548, opened 2022-09-21
Updated Ko_Kr

Kardane:
Great!


vectorwing:
Looks good! Thanks for the update.
One quick detail: you added a Mangrove Cabinet lang, but this is targeting 1.18.2, which is before it was added. I can move the lang to the 1.19 port later. 👍

PR #552, opened 2022-09-22
Update support for Botany Pots.

Darkhax:
There were a few changes to the BotanyPots schema so the existing crops would fail to load and the soils would have inverted growth rates. This PR fixes those issues and adds support for the mushroom colonies.


smbarbour:
For anyone waiting for this to get merged in:  Since these are just regular datapack json files, if you are feeling adventurous, you can replace the files in the jar with these changes.

vectorwing:
Looks good! Thanks for making the PR for these schema changes, much appreciated. 👍

HalcyonSeas73:
Are there plans to port this to 1.19.x as well, @vectorwing ?

Avakining:
A 1.19.x port would be much appreciated.

PR #561, opened 2022-10-07
Update pl_pl.json

KarolOfGutovo:
Tall-mato

Everything is translated now.
Tested


KarolOfGutovo:
I thought the patch 1 thing was a problem, but after checking a bit I think it is just github being quirky? It shows no other difference than just the pl_pl.json. I'm not sure how to go about sending the file to multiple versions so I'll only merge to 1.18.2

vectorwing:
Thanks for the lang update! 😄
Don't worry about 1.19, I can just port this lang over there. The only new line in 1.19 is for the Mangrove Cabinet, so if you'd like, you can share the line here and I'll add it manually later.

KarolOfGutovo:
The translation for mangrove cabinets would be "Namorzynowa szafka".
P.S. this mod is awesome

PR #563, opened 2022-10-10
Fire an event when eating from PieBlock for Spice of Life food tracking

### LINKED ISSUE #556: [1.18.2+ mod compat request] Spice of Life: Carrot & Sweet Potato compatibility
Hello, thank you for the mod! It is, indeed, delightful.

I'd like to request that you fire a `LivingEntityUseItemEvent.Finish` each time a player successfully eats directly from a placed food block (PieBlock), but with the item on the event being the corresponding slice rather than the whole pie item. So for example, each time a player successfully eats from a placed Apple Pie block, there would be a `LivingEntityUseItemEvent.Finish` fired that specifies the Player and the Apple Pie Slice item.

This would improve compatibility with "Spice of Life: ___ Edition" mods, which subscribe to that event in order to track which foods each player has eaten. (Sweet Potato Edition [here](https://github.com/tarinoita/Spice-of-Life-Sweet-Potato/blob/1.18/src/main/java/com/tarinoita/solsweetpotato/tracking/FoodTracker.java#L17), Carrot Edition [here](https://github.com/Cazsius/Spice-of-Life-Carrot-Edition/blob/1.18/src/main/java/com/cazsius/solcarrot/tracking/FoodTracker.java#L25).) Currently, neither one will recognize food eaten from a placed block at all, including vanilla Cake, but I plan to change this soon for Sweet Potato Edition.

In the meantime, I believe I can add compatibility within Sweet Potato by listening for the `PlayerInteractEvent.RightClickBlock` and independently computing whether the player can eat from the block or not, but I think the event makes the most sense long-term. Or if you have a better idea, I'm open to suggestions!

I've specified 1.18.2+ because that's all I really care about, but Carrot Edition goes way back, so compatibility could still be worthwhile for earlier versions of FD, if you're up for it.

Thanks for considering!

tarinoita:
Fixes #556 . Please feel free to edit or reject as you like. Thanks!

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/PieBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/PieBlock.java
@@ -29,6 +29,7 @@ import net.minecraft.world.level.pathfinder.PathComputationType;
 import net.minecraft.world.phys.BlockHitResult;
 import net.minecraft.world.phys.shapes.CollisionContext;
 import net.minecraft.world.phys.shapes.VoxelShape;
+import net.minecraftforge.event.ForgeEventFactory;
 import vectorwing.farmersdelight.common.tag.ModTags;
 
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


vectorwing:
As stated on #564, I'll be closing this PR, and pointing the other one to the 1.20 branch.

PR #566, opened 2022-10-12
ReAdd CraftTweaker Support for 1.18.2

Witixin1512:
This PR re-adds CraftTweaker support for the 1.18.2 version of Farmers Delight.
I will PR the necessary documentation to the CraftTweaker documentation repository once this is merged. I'm also open to adding support to the 1.19 branch now that CraftTweaker has released.

This PR:
- Exposes `CookingPotRecipeBookTab` to CraftTweaker
- Adds the CookingPot and Cutting Board Managers and Recipe Handlers
- Adds simple example files that showcase how to use the managers.
- Adds the 1.18.2 AnnotationProcessor to generate docs.


vectorwing:
Sorry for the late reply, @Witixin1512! And thanks for this update! I had commented those out during a version port, and forgot to reenable them.

Quick question: I see a few test scripts in the `resources` path. Are those examples going to be processed if CT is added to production, or are they ignored (used only as examples)?

build.gradle
```diff
@@ -123,20 +123,17 @@ dependencies {
     compileOnly fg.deobf("squeek.appleskin:appleskin-forge:${appleskin_version}:api")
     runtimeOnly fg.deobf("squeek.appleskin:appleskin-forge:${appleskin_version}")
 
-    // implementation fg.deobf("com.blamejared.crafttweaker:CraftTweaker-${mc_version}:${crafttweaker_version}")
+    def ctDep = fg.deobf('com.blamejared.crafttweaker:CraftTweaker-forge-1.18.2:9.1.192')
```
vectorwing: Might be better to edit and use the property values in `gradle.properties` for these lines.

Witixin1512:
The scripts inside the `data/crafttweaker/scripts/modid` only appear in the users script folder when they execute /ct examples.

That command gathers all scripts and puts them in the scripts folder for users to view and tinker with examples. On the next reload, yes, they will be executed.

src/main/java/vectorwing/farmersdelight/integration/crafttweaker/MCToolIngredient.java
```diff
@@ -1,97 +0,0 @@
-//package vectorwing.farmersdelight.integration.crafttweaker;
```
vectorwing, line 1: This question is a bit weird, since I wasn't involved in writing this class:
I noticed that `MCToolIngredient.java` was just removed, not moved to another package. Is it no longer needed for writing tool ingredients?

PR #571, opened 2022-10-26
KR translation update

Kardane:
Fixed old translation


vectorwing:
Thanks for the update!

PR #573, opened 2022-11-02
Completly re-translate Spanish Mx (es_mx.json)

cerealconjugo:
I tried to correctly adapt and translate the entire mod from scratch. I did my best to keep it loyal to default FD en_us but at the same time feel as natural as possible in Spanish Mx.

Love the mod. Thanks and i hope it helps!


vectorwing:
Thanks for updating the Mexican translation! 👍

PR #577, opened 2022-11-09
Fix traditional chinese translate

xMikux:


vectorwing:
Hey there! Thanks for the update. 😄

I noticed some of your changes seem to revert some lines which were PR'd in by a previous contributor a few days ago; as in, you're bringing them back to what they used to be. I'm unfamiliar with Chinese, so I'm not sure which of these is correct, but could you explain those changes? Just to be sure.

StarskyXIII:
I'm a translator which have  100+ mod translations experience.
I can confirm these change will be better for Taiwanese players, but still need to polish it a little bit.
I have no idea why previous contributor make a mistake on simply words.

xMikux:
Some line revert because the previous contributor has some mistake or wrong translate.

Like ``Chicken Sandwich``, correct is ``雞肉三明治``, but don't know why previous contributor change to ``肌肉三明治`` ``Muscle Sandwich``

``Flint Knife`` correct is ``燧石刀``, wrong is ``碎石刀`` ``Crushed Stone Knife``

``Iron Knife`` correct is ``鐵刀``, wrong is ``鐵斧`` ``Iron Axe``

vectorwing:
Ahh, I see! Good to know, then; I'll take one more look and then merge it. Thanks for explaining! 👍

PR #583, opened 2022-11-14
1.19.2 CraftTweaker update

Witixin1512:
This Pull Request adds CraftTweaker support to Farmer's Delight 1.19 branch.
In the process though, the mod has been updated to 1.19.2.

The mod now compiles and runs fine, but I have not tested anything inside of the mod YET.

CraftTweaker support *seems* to be working fine as of now. I do need to test replacers before giving it the full green light.
Feel free to edit upon this branch for the 1.19.2 Update.


src/main/java/vectorwing/farmersdelight/common/crafting/CuttingBoardRecipe.java
```diff
@@ -68,10 +69,11 @@ public NonNullList<Ingredient> getIngredients() {
 		return nonnulllist;
 	}
 
+	//This method is unused in regular F'D. For CraftTweaker support, if the tool goes first we can easily wrap/unwrap to and from IIngredient
```
vectorwing, line 72: This method was previously used in the JEI plugin to establish both ingredients and tool as inputs for recipe lookup, but the new JEI API defines those in a different manner.

In usual circumstances, you would be reminding me to delete an unused method, but if it's useful for CT integration, we can keep it!

vectorwing:
Well, it works locally, and everything seems fine. We can proceed with the merge. 👍

PR #586, opened 2022-11-18
Fixed some wording

ghost:
Fixed: skillet, sandy_shurb


vectorwing:
Thanks for the update!

CKenJa:
Hello!
I would like to suggest reverting this commit.
Because, the original word is intentional, not a mistake.

スキレット(skillet) -> フライパン(frying pan)
In Japan, "frying pan" is more commonly used, so I think either term is acceptable. However, the word, "スキレット," which has the same meaning as the English word "skillet." It is mainly used in outdoor cooking.

低木(low tree, bush, shrub) -> 草(grass, herb)
More appropriate are "低木".

PR #589, opened 2022-11-21
Update es_mx.json

cerealconjugo:
fixed some inconsistencies


Witixin1512 (commented):
Would you say "Cortes de Pollo" or "Tacos/Trozos de Pollo" ?

Also, "Tomate Podrido disparado" or "Tomate Podrido arrojado" ? "Disparado" feels weird without context imo...

Otherwise, this looks fine

cerealconjugo:
i think "Cortes de pollo"  fits better in this case, but i think you're right about  "Tomate Podrido Arrojado", seems to works better without context, let me change that.

cerealconjugo:
Done, thanks!

src/main/resources/assets/farmersdelight/lang/es_mx.json
```diff
@@ -274,7 +274,7 @@
   "farmersdelight.block.skillet.invalid_item": "Esto no se puede cocinar...",
   "farmersdelight.item.skillet.how_to_cook": "¡Usa tu otra mano para sostener ingredientes!",
 
-  "death.attack.farmersdelight.stove": "%1$s fue cocinado a la perfección",
+  "death.attack.farmersdelight.stove": "%1$s Fue cocinado a la perfección",
```
vectorwing: This is not impeditive, just a curiosity:
The `%1$s` represents the victim's name, making it a sentence. Wouldn't "fue" be lower-case here? Or is that how death messages are written in es_mx?

cerealconjugo: No, youre right, they mean to be lower-case. I did'nt remember why i changed that exact sentence. let me revert that.

cerealconjugo: Done:)

vectorwing:
Thank you for the updates! I'll backport the changes to 1.18.2 as well. 👍

PR #597, opened 2022-11-27
Add Botany Pots Tiers Support

PhweeDev:
Greetings. This pull request will be a simple one ahead of the 1.18 update. All this does is add support to the mod Botany Pots Tiers. It has been tested and works properly.


vectorwing:
Hi there!

I'm currently removing FD's local Botany Pots integration, as it's now located within Botany Pots itself. This integration might make things irregular or difficult to maintain with the official one, so I'm not sure if I can add it in at this point. :(

My suggestion is to propose this integration to Botany Pots Tiers itself, as it might be better suited to balance it out.

PR #605, opened 2022-12-13
Fix Bug

Umpaz:
Fix bug where if you started cooking an item, and then used the recipe book to switch the items out the cooking progress would remain


vectorwing:
Hmm, I tried testing this locally, and it didn't fix the issue you described. :(

`didInventoryChange` applies for certain actions only (which might be a design flaw), so instead of the PR fixing this, it was causing another bug:

1. Cook something and leave it in the meal slot;
2. Cook it again, and serve the stored meal during cook progress;
3. Cook time is reset wrongly.

However, adding a check based on the `checkNewRecipe` boolean (which triggers when input slots change) allows the BE to detect whether the recipe properly changed, triggering for the recipe book's quick-swap. It also would trigger if swapping two recipes with a single-ingredient difference, which has a similar scenario.

Thank you for the help, though! Didn't even notice this was bugged. 😅

PR #632, opened 2023-02-06
Fix candle cakes not dropping slices

### LINKED ISSUE #631: [MC Version]:  1.19.2
### Description

Breaking a candled cake wit a knife doesn't return any slices

### Steps to reproduce

Place cake
place candle on cake
break with knife

it doesn't give you slices

### Mod list

Farmer delight

### Logs

_No response_

### Forge

- [X] I am using the Forge version of Farmer's Delight

### Minimal instance

- [X] I have tested this on a minimal instance

### Performance and shader mods

- [X] I have tested this without performance or shader mods

edgarogh:
This PR adds cakes with candles to `loot_modifiers/slicing_cake.json`.

Fixes #631


vectorwing:
Thanks for the fix proposal! ^^

Given that Candle Cakes cannot have "slices" in the normal sense (the candle pops off and it becomes a normal cake), there's no need to bring them into the `pastry_slicing` modifier. We could probably just use `add_item` instead.

I'll test it out, then get back to you.

vectorwing:
Yup, `add_item` worked. I sent the fix just now, but thank you for your time regardless!

PR #633, opened 2023-02-06
Fix hand-held skillet ingredient rendering

### LINKED ISSUE #515: [1.19] Skillet does not display ingredient when cooking handheld
**Description**
When cooking an item on a handheld Skillet, the item is not displayed on its surface.

During the 1.19 port, the current model implementation for rendering food in a handheld Skillet has stopped working, and must be revisited. However, I was unable to do this im a timely manner, so the port will be released with this issue for now.

**I'm looking for help in regards to fixing this issue.** If anyone is looking to solve it, feel free to contact me or open a pull request.

image

**Steps to reproduce:**
1. Hold a Skillet in your hand, and a cookable ingredient in the other;
2. Stand next to a heat source;
3. Hold the use key to cook the item.

**Expected behavior:**
The Skillet should render the ingredient inside of its model, similar to the 1.18 versions.

**Mod list:**
FarmersDelight-1.19-1.1.2
Forge 41.1.0+

edgarogh:
_The diff is a joke, really._

fixes #515

---

Story time:

So basically I don't really know if what I did is clean or not. I simply did a good 30min Java debugger session to understand the different steps involved in rendering items (especially the skillet), and found out that `CompositeBakedModel::getQuads` was never called. That's an issue because there's no other way the Minecraft Item renderer could possibly know how to draw this dynamic skillet model. It needs to call `CompositeBakedModel::getQuads` at some point.

What the debugger taught me is that when rendering items, a double for loop iterates over `BakedModel::getRenderPasses` instead of directly calling `getQuads`[^note1]. And `BakedModel::getRenderPasses` returns an array of BakedModel. And its these secondary baked models whose quads are actually rendered (that makes no sense to me).

My fix, as you can see, is to just return `List.of(this)` inside `getRenderPasses`. I don't know if that's safe or if that will infinitely recurse one day. Again, I don't fully understand the logic behind nesting models.

At least the fix works, and I can't immediately think of any edge case where it would break, so I don't feel _too_ shameful about submitting code I don't fully understand. Maybe someone here who knows more about model baking will be able to sort this out?

[^note1]: In `ItemRenderer::render`, L131

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/client/model/SkilletModel.java
+++ b/src/main/java/vectorwing/farmersdelight/client/model/SkilletModel.java
@@ -17,11 +17,14 @@ import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraftforge.client.model.SimpleModelState;
+import net.minecraftforge.client.model.data.ModelData;
 import net.minecraftforge.registries.ForgeRegistries;
 import vectorwing.farmersdelight.FarmersDelight;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import vectorwing.farmersdelight.common.registry.ModItems;
+
 import java.util.*;
 
 /**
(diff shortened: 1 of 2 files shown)


RaymondBlaze:
> I don't know if that's safe or if that will infinitely recurse one day.

I believe there won't be an infinitely recurse, since you're returning `List.of(this)`, a `CompositeBakedModel` instance, then the `ItemRenderer` will only call `CompositeBakedModel::getQuads` which returns some already generated quads. So, no further method calls, no infinite recurses.

There's only two very small thing more that could be optinally optimized:
1. Try make `List.of(this)` a field so we don't have to instantize a list on every call (I know the original default method is calling `List.of(self())`, but that's basically because it's in an interface).
2. Try update all the direct calls to `BakedModel::getQuads` in the constructor to the new logic using `BakedModel::getRenderPasses`, in most cases this won't change anything, as ingredients for cooking usually won't have special models.

Anyways, good job.

edgarogh:
Thanks for the review. I'm currently away from my PC that has enough RAM to run a Minecraft modding environment but I'll gladly add another commit at the end of the week.

edgarogh:
I actually managed to run a Forge environment under 8Gb of RAM by constantly killing the daemon after each `runClient`. These commits should fix what you suggested.

---

However, I also had the idea of simply using this "render pass" system to our advantage by concatenating the skillet's render pass with the (translated-rotated) ingredient's pass**es**. This simplifies the code a bit. Again, I'm not 100% sure about the corectness of the code but it looks fine at first glance, and most importantly, works.

Here's this alternative implementation: [[tree]](https://github.com/edgarogh/FarmersDelight/tree/1.19-dev/fix-skillet-item-render-v2) [[diff with `origin/1.19`]](https://github.com/vectorwing/FarmersDelight/compare/1.19...edgarogh:FarmersDelight:1.19-dev/fix-skillet-item-render-v2)

If the alternative implementation is actually better than the one in this PR, I'll merge it into the PR's branch. A little squash might be a good idea at the end anyway.

**edit:** a few notes:
  * `ingredientBaked.getRenderPasses` takes a boolean "fabulous" argument. I don't know what it does so I set it to false
  * According to the `@Nullable` annotation, `ingredientBaked` may be null. So I'm handling the case by simply rendering only the skillet. It's better than a crash I guess.

vectorwing:
I'm back from my hiatus. Yup, just tested those fixes and they work like a charm!

image

This particular part of the code is one I don't fully understand either, but your explanation over the fix makes sense. Excellent work figuring out this missing call. Since this was peer-reviewed by @LimonBlaze, I'm willing to release this first fix on 1.2.1 right now, but we could take a look at the alternative fix some time in the future, if it would work better.

PR #650, opened 2023-02-26
Added missing translations and made corrections to Catalan localization

DyaMetR:
Added missing translations probably from the language file being outdated. It also includes grammatical and spelling corrections.


vectorwing:
Thank you for the update! 👍

PR #652, opened 2023-03-01
Datapack support for new recipies

Elaski6573:
Farmer's Delight has special recipies like cutting board.
I think new recipies can be added simply using datapack.

Datapack > data > *modid > item > *itemid.json

{ "override" : true,
      "cutting_board": {
               "reqitem": [
                        "#forge:knives" 
                ],     
               "itemresult": [
                         "**modid:**itemid", "exampleid:example"
                ]
      },
}


vectorwing:
This is the wrong panel to open issues or discussions. Please don't open a pull request if you don't know what you're doing.

Although recipes could probably be done with vanilla datapacks, I'm not interested in modifying the current system, as it would complicate compatibility with many other mods and modpacks.

PR #664, opened 2023-03-22
Implemented RecipeBuilder for Cooking Pot and Cutting Board

Lance5057:
This allows CookingPotRecipeBuilder and CuttingBoardRecipeBuilder to be used with ConditionalRecipes and other Forge data generators.


vectorwing:
Hi! Sorry for the very long delay; this fell under my radar during my hiatus. 😓

The changes here look good enough. Although FD does not use the extra methods, I can see how this could come in handy for add-ons.

I'll try to point this branch to `1.20` and see if the changes remain the same, so we can merge. 👍

PR #706, opened 2023-04-28
Added configuration option for hitboxes for some crops 

ColinGBos:
Hitboxes for the crops (tomato, cabbage, and onion) closer follow the rendering of the crops. Config option defaults to false.


vectorwing:
Hmm... I'll admit, I don't see much use for this. FD's crops follow the hitbox sizes of vanilla crops, for the most part, so this could end up inconsistent, or make it harder to harvest the crop.

Is this compatible with some other mod which changes other crops to fit such size?

vectorwing:
Sorry, but I'll have to reject it. Thanks for the proposal, though! 👍

PR #710, opened 2023-05-01
Make onions compatible with Tinkers' Construct Harvest Modifier

MegaPiggy:
The harvest modifier in Tinkers' Construct breaks and replants crops.
However, it doesn't replant onions because they have no seed like potatoes and carrots.
So we just add onions to the seeds tag like TConstruct does for potatoes and carrots.


vectorwing:
Though I understand that TCon uses this to determine harvesting targets... it kinda seems like a wrong tagging.

I imagined that `forge:seeds` was for actual seed items, since it lacks Carrot and Potato by default. Not sure what the impact of adding Onions would be. I'll ask around a bit, then get back to you.

vectorwing:
After looking a bit further, it seems TCon does _not_ use `forge:seeds`, but instead makes its own `tconstruct:seeds` tag which includes direct-plantables.

I'll add Onion to this tag locally for compatibility. Thanks for letting me know about this, though! ^^

Issue #712, opened 2023-05-04
[Suggestion] Add fruits/vegetables and plants to Create: Crafts & Additions tags


saltyseadoggo:
As per the title, I am requesting that Farmer's Delight add its various fruits and vegetables to `#createaddition:plant_foods`, and its various plant items to `#createaddition:plants`. This will allow FD items to be used to make C:C&A biomass, without players/modpack devs needing to add all these items to these tags themselves.~
More specifically...

`#createaddition:plant_foods` should contain
`farmersdelight:cabbage`
`farmersdelight:tomato`
`farmersdelight:onion`
`farmersdelight:pumpkin_slice`

`#createaddition:plants` should contain
`farmersdelight:sandy_shrub`
`farmersdelight:wild_cabbages`
`farmersdelight:wild_onions`
`farmersdelight:wild_tomatoes`
`farmersdelight:wild_carrots`
`farmersdelight:wild_potatoes`
`farmersdelight:wild_beetroots`
`farmersdelight:wild_rice`
`farmersdelight:brown_mushroom_colony`
`farmersdelight:red_mushroom_colony`
`farmersdelight:rice_panicle`

vectorwing:
I took a look at how C:C&A is handling its Biomass recipes, and it seems some of those are already covered in some recipes:
- The recipe asking for `forge:crops` accepts all FD crops, except for the Rice Panicle;
- The recipe asking for `minecraft:flowers` accepts all Wild Crops, as they're tagged as such;

This would leave a few of those out still, so we could still add this tag for them. I can take a look later. 👍

PR #719, opened 2023-05-21
added tag compat for create crafts and additions

moltenwolfcub:
Implemented the tags requested in issue #712.


vectorwing:
Heyo, @moltenwolfcub, thanks for the PR! ^^

After checking the add-on, I noticed many of those items are already accounted for in different tags. Only 6 of them would require direct tagging, as seen below:

```
plants:
"farmersdelight:sandy_shrub"
"farmersdelight:brown_mushroom_colony",
"farmersdelight:red_mushroom_colony",

plant_foods:
"farmersdelight:pumpkin_slice"
"farmersdelight:rotten_tomato"
"farmersdelight:rice_panicle"
```

Also, I have recently added compatibility tags to data generation. Could you create these tags through there, so they can be generated together with `runData`? I can do this locally instead, if you prefer. 👍

Here's the file:
https://github.com/vectorwing/FarmersDelight/blob/1.18.2/src/main/java/vectorwing/farmersdelight/common/tag/CompatibilityTags.java

moltenwolfcub:
I'll have a look at it.

moltenwolfcub:
I think that should be updated to use datagen and those updated lists of items to be used in the tags.

vectorwing:
Perfect! Thank you for the update; should be out on 1.2.2 soon. 👍

PR #731, opened 2023-06-04
Improved Integration support for custom recipes

ColonelPanic-Dev:
Hi Vectorwing,
these changes are intended to allow a bit more integration support to your recipes. This is done by calling assemble() like vanilla recipes do, instead of getResultStack(). That way a custom recipe can have access to the state of the inventory at the time of crafting. Obviously this isn't appropriate for the cutting board, so instead I just added a RecipeWrapper to the rollResults method signature. 

Due to these changes, Item checks have to be expanded a little bit to include comparing nbt tags, to prevent recipes that apply nbt tags having improperly stacking results.

All-in-all its a very small change that brings the recipes a little more in line with vanilla implementation as well as providing a lot more information to the recipe implementation for modders to have access to


src/main/java/vectorwing/farmersdelight/common/crafting/ingredient/ChanceResult.java
```diff
@@ -77,7 +77,7 @@ public static ChanceResult deserialize(JsonElement je) {
 		float chance = GsonHelper.getAsFloat(json, "chance", 1);
 		ItemStack itemstack = new ItemStack(ForgeRegistries.ITEMS.getValue(new ResourceLocation(itemId)), count);
 
-		if (GsonHelper.isValidPrimitive(json, "nbt")) {
+		if (json.has("nbt")) {
```
ColonelPanic-Dev, line 80: Best I can tell, an nbt is not a JsonPrimitive and so Gson.isValidPrimitive never evaluates to true even if the json is properly formatted. I changed it to just check if the nbt is present which I think is sufficient and in line with how forge's Crafting Helper does it

vectorwing:
Hello! thank you for the PR, and sorry for the late reply; I usually reserve recipe spec changes for minor releases.

Yeah, everything here seems accurate enough. Using `assemble()` would let extenders of `CookingPotRecipe` do something with the inventory, and `isSameItemSameTags` allows NBT results to be correctly merged.

I'm pointing this PR to `update/1.3`, to organize those changes in one place before release. 👍

src/main/java/vectorwing/farmersdelight/common/crafting/ingredient/ChanceResult.java
vectorwing: Yeah... not sure why it's calling this method. While much of ChanceResult came from Create, they never used this call, apparently.

I might change it to `GsonHelper.isValidNode()` instead, but for now, this one should suffice.

vectorwing:
Hi! I haven't forgotten about this! 😆
I pointed it to `dev/1.3`, to include it amidst other minor changes for 1.20, as I have now abandoned 1.19.

PR #733, opened 2023-06-05
Add non object ingredient support for cutting recipes

BobVarioa:
This essentially is just Forge [CompoundIngredient](https://docs.minecraftforge.net/en/1.18.x/resources/server/recipes/ingredients/#compoundingredient) support, does not change functionality for anything else in the mod, and all current recipes are compatible.


vectorwing:
Hmm, good catch. I didn't realize `tool` wasn't compatible with array ingredients.

However, after digging through some vanilla code, I feel like checking specifically for a `JsonArray` might be a safer approach than pulling a `json.get()` directly. I'll add the line manually for 1.20.1 and above.

I'll be closing this PR, as I'm no longer supporting 1.18. Thanks for taking the time to PR, though! Apologies for the long delay. ^^

vectorwing:
Added on commit dcf0ac4. 👍

PR #757, opened 2023-07-07
An option that lets Deployers not place items on Cutting Boards

JerryLu086:
A new option that determines if Create Deployers should be able to place items on Cutting Boards or not (false by default). This makes automating Cutting Board recipes easier (or actually just makes it able to really automate them?).

As tested:
- Using the same ingrediants for cutting is about 13.3% faster than deploying
- Axes on the boards are basically the same speed as Mechanical Saw cutting, while they require tools and cost durability

Due to how Create's new code was written to prevent Deployers from not working in protected chunks, so far it's using the display name to detect if the player entity using the board is a Deployer or not. Thanks!

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/Configuration.java
+++ b/src/main/java/vectorwing/farmersdelight/common/Configuration.java
@@ -21,6 +21,7 @@ public class Configuration
 	public static ForgeConfigSpec.DoubleValue CUTTING_BOARD_FORTUNE_BONUS;
 	public static ForgeConfigSpec.BooleanValue ENABLE_ROPE_REELING;
 	public static ForgeConfigSpec.ConfigValue<List<? extends String>> CANVAS_SIGN_DARK_BACKGROUND_LIST;
+	public static ForgeConfigSpec.BooleanValue DEPLOYER_PLACES_ITEM_ON_CUTTING_BOARD;
 
 	public static final String CATEGORY_FARMING = "farming";
 	public static ForgeConfigSpec.ConfigValue<String> DEFAULT_TOMATO_VINE_ROPE;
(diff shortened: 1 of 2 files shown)


vectorwing:
Heyo! Thanks for the PR.

I'm not sure if something like this is needed, to be honest. Yes, the Deployer can keep placing and removing the tool repeatedly, but so could any other fake player from other mods. This is a behavior that I'm still looking into in regards to the Cutting Board, and would rather find a more global solution for it.

You can circumvent this behavior, or "solve the puzzle", with basic redstone and Create, as shown below:

image

Cutting Boards output a maximum comparator signal when holding an item. If the signal is reversed, a clutch can be triggered whenever the Cutting Board is out of items, preventing the Deployer from ever placing its tool on it.

Please let me know what you think!

JerryLu086:
Was playing with A&B and ran into the Slimy Fern recipes and found it a bit inconsistent, haven't touched vanilla redstone for years (I've thought of using them to do this though, just thought it could be a little nice addition for the mod). 

Seems like it could be hard, since we don't even have a common way (if not touching other mods' code directly) to get info that we can tell if it is a Fake Player and which it is. The simplest way (but the most annoying to do) is to add the support for Fake Player blocks from popular mods one by one, but yeah, it's not a good idea to go for.

 Maybe just leave this for now, since everybody except me might have already been using this method for years now lol. Thanks for responding \:)

vectorwing:
I'll admit that I'm not sure if most players reach the same conclusion as I did above when working with this setup, but given it's possible to "control" the Deployer's interactions this way, I'd chalk it up to part of the puzzle for now.

There is a chance the Cutting Board will receive some reworks later, which could help make those interactions more stable, but that will be done later down.

Sorry, but I'll have to reject this proposal. Thanks for the help, either way! ^^

PR #763, opened 2023-07-12
update pl_pl.json

jogurciQ:
Added hanging canvas signs, cherry and bamboo cabinets and fixed many gramatical errors


vectorwing:
Thank you for the update! ^^

PR #781, opened 2023-07-21
Update es_es.json

ghost:


Witixin1512 (commented):
- You refer to the Skillet in most places as `Sartén`, yet the default translation is `Cocina Portátil`. Why is that?
- `¡Detente cerca del calor y sostenga` - Why the change in tone here? Also, shouldn't it be stand on top of heat?

ghost:
> * You refer to the Skillet in most places as `Sartén`, yet the default translation is `Cocina Portátil`. Why is that? 

It's not, the default translation is Sartén (Line 6). Cocina Portátil is the advancement name.
> * `¡Detente cerca del calor y sostenga` - Why the change in tone here? Also, shouldn't it be stand on top of heat?

The change in tone is due to there being no literal translation for "Stand near heat". Also,  the default text is "Stand near heat" not "Stand on top of heat".

PR #794, opened 2023-08-10
Add localized name for the Rotten Tomato Entity

leagris:
The Rotten Tomato Entity localized name is shown when a Rotten Tomato flies and is pointed by player cursor with mods like Waila.

This patch adds the missing localized name for the Rotten Tomato entity, to every languages with a Rotten Tomato Item translation key, as it reuses the same translation for the Entity.


vectorwing:
I see that cursor tooltip mods continue to introduce use cases for new langs. 😅

I was checking your french translation PR before this one, and it confused me as to why none of my langs had this line anywhere. I'll merge this PR after the other translation PR's, so that there's only 1 conflict source to resolve; though I may move the entity category lower down later.

Thank you for the update! 👍

PR #798, opened 2023-08-16
Update ru_ru.json

VGamerGroup:
# Update for Russian Localization for 1.20 and 1.20.1

## About update

- Added translation for hanging signs **ALL** colors
- Added translation for mangrove, cherry and bamboo cabinets

##### Thanks for creating your mods for us!


vectorwing:
Thank you for the update!

PR #804, opened 2023-08-29
Subclassable Recipe Builders

Lordfirespeed:
When an addon mod for Farmer's Delight (Farmer's Respite, Nether's Delight, etc) want to register a `CuttingBoardRecipe` or `CookingPotRecipe`, they cannot subclass the existing recipe builders as their constructors are private. That means developers are limited to 
- ~~making an access transformer that modifies the recipe builders' constructor~~ (access transformers do not work on other mods)
- copy + pasting the implementation of the recipe builder (present solution 😭)

This PR enables subclassing the recipe builders, for example
```java
package umpaz.nethersdelight.data.builder;

import net.minecraft.MethodsReturnNonnullByDefault;
import net.minecraft.resources.ResourceLocation;
import net.minecraft.world.level.ItemLike;

import javax.annotation.Nullable;
import javax.annotation.ParametersAreNonnullByDefault;

@ParametersAreNonnullByDefault
@MethodsReturnNonnullByDefault
class NethersDelightCookingPotRecipeBuilder extends CookingPotRecipeBuilder {
    protected NethersDelightCookingPotRecipeBuilder(ItemLike resultIn, int count, int cookingTime, float experience, @Nullable ItemLike container) {
        super(resultIn, count, cookingTime, experience, container);
    }

    public static NethersDelightCookingPotRecipeBuilder cookingPotRecipe(ItemLike mainResult, int count, int cookingTime, float experience) {
        return new NethersDelightCookingPotRecipeBuilder(mainResult, count, cookingTime, experience, null);
    }

    public static NethersDelightCookingPotRecipeBuilder cookingPotRecipe(ItemLike mainResult, int count, int cookingTime, float experience, ItemLike container) {
        return new NethersDelightCookingPotRecipeBuilder(mainResult, count, cookingTime, experience, container);
    }

    @Override
    protected String getDefaultRecipeName(ResourceLocation resultItemKey) {
        return "nethersdelight" + ":cooking/" + resultItemKey.getPath();
    }
}
```

This is achieved by making the constructors `protected` and providing an overridable method `getDefaultRecipeName` on each RecipeBuilder.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/data/builder/CookingPotRecipeBuilder.java
+++ b/src/main/java/vectorwing/farmersdelight/data/builder/CookingPotRecipeBuilder.java
@@ -103,9 +103,14 @@ public class CookingPotRecipeBuilder
 		return this;
 	}
 
+	protected String getDefaultRecipeName(ResourceLocation resultItemKey) {
+		return FarmersDelight.MODID + ":cooking/" + resultItemKey.getPath();
+	}
+
 	public void build(Consumer<FinishedRecipe> consumerIn) {
 		ResourceLocation location = ForgeRegistries.ITEMS.getKey(result);
-		build(consumerIn, FarmersDelight.MODID + ":cooking/" + location.getPath());
+		assert location != null;
+		build(consumerIn, getDefaultRecipeName(location));
 	}
 
 	public void build(Consumer<FinishedRecipe> consumerIn, String save) {
--- a/src/main/java/vectorwing/farmersdelight/data/builder/CookingPotRecipeBuilder.java
+++ b/src/main/java/vectorwing/farmersdelight/data/builder/CookingPotRecipeBuilder.java
@@ -41,7 +41,7 @@ public class CookingPotRecipeBuilder
 	private final Item container;
 	private final Advancement.Builder advancement = Advancement.Builder.advancement();
 
-	private CookingPotRecipeBuilder(ItemLike resultIn, int count, int cookingTime, float experience, @Nullable ItemLike container) {
+	protected CookingPotRecipeBuilder(ItemLike resultIn, int count, int cookingTime, float experience, @Nullable ItemLike container) {
 		this.result = resultIn.asItem();
 		this.count = count;
 		this.cookingTime = cookingTime;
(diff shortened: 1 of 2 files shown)


vectorwing:
Hello! Sorry for the long delay, @Lordfirespeed. ^^

I suppose I could un-private the constructors for the builders. When they were written, I didn't consider that add-ons would be using them internally.

That said, what is the use case here? In your example, it seems you're simply using the builder without changes to its format; you can use different overrides for the `.save()` method to define specific namespaces and path names. Is the goal to simplify this process, or to accomodate a custom cooking recipe, with different parameters than vanilla FD?

Lordfirespeed:
Cor blimey, it's been a while on this - I remember I was working on [nether's delight](https://github.com/Chefs-Delight/NethersDelight_Forge/blob/1.20.1/src/main/java/umpaz/nethersdelight/data/builder/NDCookingPotRecipeBuilder.java)

I think the goal was to simplify the process - I'm not 100% sure about that

vectorwing:
Hello again! Sorry for another long delay, I wasn't able to decide on this PR at the time.

I decided to sit down and understand what's going on, as other add-on developers also expressed an interest in using the recipe builders. The method `save(consumer)` in `RecipeBuilder` picks the result's mod ID to choose which path to save to, requiring users to specify a manual ResourceLocation if they want a particular path and filename. Most vanilla builders use the default I described.

However, due to FD needing to specify a sub-folder, it overrides *and* hardcodes FD's own mod ID to the result, which was shortsighted of me. Ideally, it would repeat what `RecipeBuilder` does and simply append `cooking/` to the path. This would eliminate the need to define the mod ID when the result belongs to your mod, only needing manual pathing if it isn't, like what FD itself does in every such case.

What we could do here is:
- Update FD's `save(consumer)` override to get the result's key, and insert the `cooking/` subdirectory;
- Add a new `save(consumer, modid)` to simplify the usage when it comes to add-ons.

The changes above would eliminate the need to extend the recipe builder itself, if all you want is to datagen a cooking or cutting recipe.

@Lordfirespeed, I can do these changes myself if you prefer, since I took too long to follow up on your PR. But do let me know if you're still interested in tackling this, and/or if the change I proposed would be useful in your case!

vectorwing:
Update: The more I read my own implementation, the more I realize how much crap is left over from before it inherited RecipeBuilder. There's a lot to clean up...

I think I'll update them on my own, as there is plenty of things to change. Thanks for the PR anyway!

PR #807, opened 2023-09-02
Abstract the Stove so it can be subclassed 

Lordfirespeed:
This allows extension mods e.g. Nether's Delight to register their own stoves without having to copy-paste any implementation!


vectorwing:
Hello! 👋

I believe we had discussed this PR previously on Discord, so I figured I'd ask here this time: are you still interested in updating this PR for the current version?

Due to the nature of the change, I can only add it in a minor version. 1.3 is mostly finished and being ported/tested to 1.21, so let me know if you'd like to update it before this release, and I can take another look.

Lordfirespeed:
Hallo! I got started on it but it's not done yet. I ran into a few behaviour questions which slowed me down 😅

E.g. presently when an opaque block is placed on top of a stove which is not lit, items are not ejected. is this intentional?

vectorwing:
I don't think it's intentional, or at least I didn't intend for that. Probably some forgotten clause inherited from the Campfire.

Either way, this is something I can fix later down, regardless of the abstraction. 👍

Lordfirespeed:
Superceded by #1265

PR #860, opened 2024-01-06
Update zh_tw.json

ghost:


vectorwing:
Hello! Thanks for the update!
I changed the PR's target to 1.20, as the langs you added only exist there. 👍

PR #885, opened 2024-02-26
added translation for british english

FernLeafie:
gray canvas signs are now called grey canvas signs when using the british translation 
since that was missing (light grey canvas signs are also included in this)


vectorwing:
Okay, this is too funny not to add. 😆 
I'll be redirecting this PR to 1.20, as 1.19 is now abandoned. I can add lines for Hanging Canvas Signs manually after that.

PR #890, opened 2024-03-27
Skillet stuff

MehVahdJukaar:
This PR does the following:
- Add skillet flipping animation
- Makes skillet use Item Renderer instead of baked models while still keeping it json model based to easily position it with blcokbench.
- Adds skillet cook bar while cooking (replaces durability bar there)
- replaces various instances of InteractionResult.SUCCESS with the correct .sideSuccess(level.isClientSide)
- simplifies some logic around the above mentioned code.

Also i realized i accidentally worked on the 1.19 branch. This PR should be able to be cherry picked into 1.20 main aswell with minimal changes (A single argument in ISTER method changed in newer forge)


MehVahdJukaar:
In last commits I added the following:
- Replaced all RegistryObject fields with Supplier fields. This is just so multi loader addons can reference both versions registries directly. Same was done for Configuration class
- Added skillet block flipping and sound

vectorwing:
Deeply sorry for leaving this PR behind for this long. If I recall, I had asked you to extract a few things from it, but I never returned to review it.

@ChrysanthCow recently updated a newer PR with a different skillet PR implementation from Refabricated, and I ended up merging it. Either way, thank you for taking the time to make the PR, and sorry again.

PR #899, opened 2024-04-12
[Mod Integration][Botania] Make Pies edible for Kekimurus

RaymondBlaze:
### Intro
The Kekimurus in Botania is known for generating mana from eating cakes. Botania's code only checks for `CakeBlock` and thus only vanilla cake and some modded cakes.

Since FD's pies(including the sweet berry cheesecake) behaves like vanilla cake, it's reasonable for pies to have integration with Kekimurus.

It's possible to make this integration on Botania's side, but that would be requiring Botania to compile against FD. Since I found a way to implement it on FD's side without needing to compile against Botania, I believe making a PR to FD would be better.

[Related Discord discussion](https://discordapp.com/channels/734511833947439156/734514412873973830/1227887864633753620)

### Implementation
This PR used a `@Pseudo` mixin on the `KekimurusBlockEntity` to avoid adding Botania as a hard dependency on compile time.

Not having Botania on the compile classpath results in not being able to call Botania specific methods in the mixin, however with MixinExtras, a safe and possibly more coremod-compatible implementation could be made. The mixin result will basically be equivalent to the code below:
<details>

<summary>Click to uncollapse</summary>

```java
/*
 * This class is distributed as part of the Botania Mod.
 * Get the Source Code in github:
 * https://github.com/Vazkii/Botania
 *
 * Botania is Open Source and distributed under the
 * Botania License: http://botaniamod.net/license.php
 */

@Override
public void tickFlower() {
	boolean isPie = false;
	super.tickFlower();

	if (getLevel().isClientSide) {
		return;
	}

	int mana = 1800;

	if (getMaxMana() - this.getMana() >= mana && !getLevel().isClientSide && ticksExisted % 80 == 0) {
		for (int i = 0; i < RANGE * 2 + 1; i++) {
			for (int j = 0; j < RANGE * 2 + 1; j++) {
				for (int k = 0; k < RANGE * 2 + 1; k++) {
					BlockPos pos = getEffectivePos().offset(i - RANGE, j - RANGE, k - RANGE);
					BlockState state = getLevel().getBlockState(pos);
					Block block = state.getBlock();
					if ((isPie = block instanceof PieBlock) && (block instanceof CakeBlock)) {
						int nextSlicesEaten = state.getValue(isPie ? PieBlock.BITES : CakeBlock.BITES) + 1;
						if (nextSlicesEaten > (isPie ? 3 : 6)) {
							getLevel().removeBlock(pos, false);
						} else {
							getLevel().setBlockAndUpdate(pos, state.setValue(isPie ? PieBlock.BITES : CakeBlock.BITES, nextSlicesEaten));
						}

						getLevel().levelEvent(LevelEvent.PARTICLES_DESTROY_BLOCK, pos, Block.getId(state));
						//Usage of vanilla sound event: Subtitle is "Eating", generic sounds are meant to be reused.
						getLevel().playSound(null, getEffectivePos(), SoundEvents.GENERIC_EAT, SoundSource.BLOCKS, 1F, 0.5F + (float) Math.random() * 0.5F);
						addMana(mana);
						sync();
						return;
					}
				}
			}
		}
	}
}
```

</details>

### Changes
MixinExtras is now embedded in the mod's jar using Forge's jar in jar, so no extra dependency for players is introduced. Since NeoForge 20.2.84+ includes MixinExtras already, removing the embedded dependency when FD updates to NeoForge is possible.

The only change for the building workflow is that now the `jarJar` task should be used to create the jar used for production instead of the original `jar` task.

### Discussion
The Kekimurus generates 1800 mana per bite for all cakes, currently I let the pie use the same value. However, since pie slices are generally better than cake slices, and pies only have 4 bites comparing to cake's 7 bites, we might should consider a proper value for the mana generated per bite for pies.

My current preferred value is 3200 mana, which is 12800 mana in total, while for cake it's 12600 mana in total.

Of course, this value should be made tweakable through config, but we still need some discussion to determine a default value.

If the discussion results in rejecting this integration PR, I shall port it to Create: Central Kitchen for those who're interested in it :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,8 @@ mixin {
     add sourceSets.main, "farmersdelight.refmap.json"
 }
 
+jarJar.enable()
+
 version = "${mc_version}-${mod_version}"
 group = 'vectorwing.farmersdelight' // http://maven.apache.org/guides/mini/guide-naming-conventions.html
 archivesBaseName = 'FarmersDelight'
@@ -97,6 +99,7 @@ minecraft {
 }
 
 repositories {
+    mavenCentral()
     maven {
         // JEI
         name = "Progwml6 maven"
@@ -116,6 +119,11 @@ sourceSets.main.resources {
 
 dependencies {
     minecraft "net.minecraftforge:forge:${mc_version}-${forge_version}"
+
+    compileOnly(annotationProcessor("io.github.llamalad7:mixinextras-common:0.3.5"))
+    implementation(jarJar("io.github.llamalad7:mixinextras-forge:${mixinextras_version}")) {
+        jarJar.ranged(it, "[${mixinextras_version},)")
+    }
 
     // compile against the JEI API but do not include it at runtime
     compileOnly fg.deobf("mezz.jei:jei-${mc_version}-common-api:${jei_version}")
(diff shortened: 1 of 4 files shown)


vectorwing:
Hmmm... Given this changes the build workflow directly, which wouldn't be the case in NeoForge since it adds this by default, I prefer adding this PR to the **1.20.4** branch of FD instead, which is already fully ported to NeoForge. 1.20.1 will likely be discontinued soon anyway.

Can you please reopen this PR pointing to it? The build is fully functional, I just haven't finished testing it for a release. 👍

RaymondBlaze:
I was not aware of the 1.20.4 branch being fully functional, and given that Botania's newest version was 1.20.1, so I targeted the 1.20 branch ;)

I'll reopen the PR pointing the 1.20.4 branch, I would suggest leave it open until Botania is updated to 1.20.4 so we can test the integration (it worked fine on 1.20.1 and would likely to work on 1.20.4).

PR #905, opened 2024-04-21
Update de_de.json

joernott:
- Fixing a lot of capitalization errors
- Fixing a few Enghhlish/German mixups
- Fixing wrong translation of a plate of roasted chicken


joernott:
One of my players plays MC in German and stumbled upon the wrong translation of a plate of roasted chicken. He asked me to fix it in my mod pack but I#d rather submit a fix to the mod author instead. While doing so, I've stumbled upon a few capitalization errors and some bad wording, too.

So, here is the PR for those fixes.

Moralle:
Looked over this and it seems like good changes for the most part, though a few notes:

> "farmersdelight.advancement.harvest_straw.desc": "Ernte Kräuter und Feldfrüchte mit dem Küchenmesser, um Stroh zu bekommen.", 

"Kräuter" wouldn't be accurate since that mostly means "Herbs", "krautige Pflanzen" seems like the better translation even if its a bit clunkier

> "farmersdelight.advancement.place_organic_compost.desc": "Plaziere ein wenig organischen Kompost. Es kompostiert besser wenn es mit Sonne, Wasser oder Pilzen in Kontakt ist"

Small typo there that was carried over from the previous translation, it should be "Platzieren"

> "farmersdelight.block.cutting_board.invalid_item": "Das kann nicht zerschnitten   werden...",

The german is fine but there are too many spaces between "zerschnitten" und "werden"

This PR would be an improvement IMO but if these things could be adjusted that'd be even better

joernott:
Thanks for the review @Moralle. I've fixed the translations for the ones you mentioned.

vectorwing:
Thank you for the translations, and thanks @Moralle for the reviews!
I'll point the PR over to the 1.20 branch, as 1.19 is no longer being maintained.

PR #929, opened 2024-07-04
Int to Float in CuttingBoardRecipeBuilder

brnbrd:
Was impossible to have a sub-100% chance due to incorrect type


vectorwing:
Good eye! I hadn't noticed because I don't use this constructor internally, only `addResultWithChance()`. 😄

PR #895, opened 2024-04-06
Update `ru_ru.json`

mpustovoi:
Fixed typos, case and synced with `en_us.json`

PR #953, opened 2024-08-01
Update `ru_ru.json` for 1.21

mpustovoi:
Copy of https://github.com/vectorwing/FarmersDelight/pull/895, but for 1.21

### DIFF
--- a/src/main/resources/assets/farmersdelight/lang/ru_ru.json
+++ b/src/main/resources/assets/farmersdelight/lang/ru_ru.json
@@ -211,7 +211,7 @@
   "enchantment.farmersdelight.backstabbing": "Предательство",
   "enchantment.farmersdelight.backstabbing.desc": "Увеличивает урон при атаке цели со спины.",
 
-  "itemGroup.farmersdelight": "Фермерский восторг",
+  "itemGroup.farmersdelight": "Фермерский Восторг",
 
   "farmersdelight.container.recipe_book.cookable": "Доступно к готовке",
   "farmersdelight.container.cooking_pot": "Кухонный котел",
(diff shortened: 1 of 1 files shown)


TheGridExpert:
I agree with the changes, but I wonder if the name of the mod should be translated (in lines 214 and 233)? I think many people know Farmer's Delight by its original name.

mpustovoi:
> I wonder if the name of the mod should be translated

I rely on the fact that I was once told that English words don't mean anything to a small child that play this game.
And once this mod has already been downloaded, why not.
After all, on Modrinth or CurseForge, it appears under the English name.
Although, maybe you're right.
Minecraft Crowdin

TheGridExpert:
Well, mods are more often played by "older children" for which it does make sense
Or maybe that's just because I don't like to localise titles, because their translated variants sounds weird for me

vectorwing:
I usually leave this decision to the translator, since it only really affects the Advancements tab. Whichever convention fits best is okay. 👍

mpustovoi:
> Well, mods are more often played by "older children" for which it does make sense

I agree, but I was strongly proved the opposite somewhere in the `Fabulously Optimized` repository.

> Or maybe that's just because I don't like to localise titles, because their translated variants sounds weird for me

I think it depends. 
In the case of `Farmer's Delight`, it doesn't seem bad to me.
However, for example, in the case of the `Create` mod, any option will sound weird.
Though, in the case of a mod that adds one entity, for example, `Hunter's Return` or `Cave Dweller`, it is better to translate everything.

mpustovoi:
I think putting the first letters of the words in uppercase would be a bit nicer.
No idea why I didn't do it earlier.

TheGridExpert:
Well, in Russian usually only the first letter in title is capital (not considering names, places etc.), so I suppose both make sense

mpustovoi:
Yes, sure.
Let it be an exception here.

PR #957, opened 2024-08-04
Update and Backport German Translations

DaLumma:
I added translations for unterwater Skillets and I hope you don't mind that I also added all the previous updates and fixes to 1.20.


Moralle:
Some notes:

> "item.farmersdelight.squid_ink_pasta": "Pasta mit Sepia",

This is not necessarily wrong as far as I can tell, but neither would be "Tintenfischtinte". "Sepia" would be less wordy, but "Tintenfischtinte" is a little more on-the-nose

> "item.farmersdelight.honey_glazed_ham": "Eine Portion Honigglasierter Schinken",

"honigglasiert" has to either be lowercase since it's an adjective, or has to be separated into "Honig-glasiert"

> "farmersdelight.advancement.get_fd_seed.desc": "Vier neue Pflanzen können in verschiedenen Klimas gefunden werden... oder vielleicht in einer Kiste irgendwo.",

While "Klimas" has also been added to the Duden, the plural of "Klima" should be "Klimata" ("Klimate" seems also acceptable)

DaLumma:
Thanks I didn't notice the "Pasta mit Sepia" i'll leave that

vectorwing:
Thanks for backporting! 👍

PR #961, opened 2024-08-10
(1.21.1) Fix crash when placing Canvas Signs (and hanging signs)

Mrbysco:
* Update Neoforge to 1.21 and update some dependencies
* Fix crash with the Canvas sign block entity and hanging block entity


vectorwing:
Thanks for the fix! 👍

PR #966, opened 2024-08-19
Update Fabric port on Bug Report form.

ChrysanthCow:
This PR changes the Fabric port to Farmer's Delight Refabricated's repository instead of the now deprecated Farmer's Delight Fabric.
https://github.com/MehVahdJukaar/FarmersDelightRefabricated


vectorwing:
Right, good call! Thanks for the PR. 👍
Can you point it to the 1.20 branch, though? It's the current default, until I release the last FD updates for it.

PR #971, opened 2024-08-23
[1.20.1] Add EMI compat.

### LINKED ISSUE #877: From [EMI 1.1.0+1.20.1+forge] Percentaged output in recipe tree
### Minecraft version

1.20.1

### Farmer's Delight version

1.2.4

### Forge version

47.2.17

### Description

An issue from emi's page. They told me the following is a bug with Farmers Delight:
EMI indicates in the recipe that an item has a probability (here 50%) for the recipe (here cutting board) as an output. But in the recipe tree it doesn't show ~8 (4/50%) wild rice like in the fabric version.
303180673-5ead2d68-2d51-4777-a4f7-2a1a0d7f1650
303180667-0929c754-6858-4988-afd6-2b62a136541d

[https://github.com/emilyploszaj/emi/issues/432](url)

### Steps to reproduce

Open the inventory
Look for a recipe with a percentaged output like shovel-digging gravel on a cutting board (did you want the duplication-glitch?)
Press the button with the four squares next to the recipe.
Adjust the batch size. The leftovers (flint) are always as many as gravel, but the should be a tenth. The same if you wanna make a flint knife for example and choose for the cutting recipe the are not 10 gravel but 1 gravel.

### Mod list

Create 1.20.1-0.5.1.f (has also this problem)
EMI 1.1.0+1.20.1+forge
Farmer's Delight 1.20.1-1.2.4
Just Enough Items 15.3.0.1 (without it the recipes weren't visible)

### Logs

_No response_

### Minimal instance

- [X] I have tested this on a minimal instance

### Performance and shader mods

- [ ] I am using performance or shader mods

ChrysanthCow:
Resolves #877

The main difficulty here was translating every tag, but ideally, you won't have to change the tags post 1.20.1.

Look at the 1.21 PR #972 to get a general idea of how difficult updating may be. EMI has a really good API, so I'd imagine it wouldn't be too painful, but ultimately, it's your call whether this gets in or not.


Rimevel:
In case there is no desire to merge this into FD, could this be a standalone addon instead?

ChrysanthCow:
> In case there is no desire to merge this into FD, could this be a standalone addon instead?

Absolutely! I'll make the addon.

Rimevel:
Any news on a potential addon to fix this? :D Farmers Delight is the only mod in my pack that does not show recipes in the crafting menu.

ChrysanthCow:
> Any news on a potential addon to fix this? :D Farmers Delight is the only mod in my pack that does not show recipes in the crafting menu.

Ive been busy lately and i want to wait on this PR's resolution before i put effort into that.

vectorwing:
Thanks for moving the PR to the `dev/1.3` branch! 👍

I wasn't sure whether to support another item viewer natively, due to maintenance, but after reading its docs to get more context for the PR, EMI seems pretty easy to manage.

From the looks of it, the main new things to keep an eye on are:
- Default recipes (editable in-game or manually);
- Tag langs (JEI now uses them too, good standard to use moving forward).

I'll do some local testing, but everything here looks good for approval so far!

PR #974, opened 2024-08-24
Cutting Board Recipe Patch

Lance5057:
Patches cutting board recipe to allow items to have multiple outcomes based on tool used.
image
image
image
An example of the issue.


vectorwing:
Note to self: this can be safely added to 1.21 only, as this bug doesn't occur in 1.20.1.
I'll take a look at it later!

PR #976, opened 2024-08-26
Update zh_tw.json

ghost:


vectorwing:
Thank you for the update!

I'm only able to verify langs with Google Translate, but from the ones I checked, it seems some of your updates aren't as accurate as the previous ones. I might need the review of another Chinese translator here before approval.

PR #977, opened 2024-08-26
Fixed mangrove cabinet crash

theendercore:


vectorwing:
Thanks for the fix! 👍

PR #979, opened 2024-08-27
Make shapes of pies and feasts rotate depending on their direction

TheGridExpert:
Added shapes for each stage of the pies and made them rotate depending on the direction. Same with the feasts, but only for Shepherd's Pie, Honey Glazed Ham and Roast Chicken. I decided not to do this for the Rice Roll Medley because the model would be far too complicated.

https://github.com/user-attachments/assets/4b55e654-7ff1-429a-af03-ce78a31c9601

(Also, AddItemModifier, FDAddTableLootModifier and PastrySlicingModifier were made public in order to allow data generation with them in addon mods)

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/FeastBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/FeastBlock.java
@@ -23,8 +23,11 @@ import net.minecraft.world.level.block.state.properties.DirectionProperty;
 import net.minecraft.world.level.block.state.properties.IntegerProperty;
 import net.minecraft.world.level.pathfinder.PathComputationType;
 import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.shapes.BooleanOp;
 import net.minecraft.world.phys.shapes.CollisionContext;
+import net.minecraft.world.phys.shapes.Shapes;
 import net.minecraft.world.phys.shapes.VoxelShape;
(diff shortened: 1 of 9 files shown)


vectorwing:
Thanks for the contribution!

However, the way this is done on your PR might have significant performance problems. VoxelShapes should not be computed on the fly like this, as joining operations are expensive. It's sort of why I didn't do it originally. 🙁

I asked for some pointers in regards to this on Discord, and it seems this could be executed if these shapes were lazily computed and cached as they're called. In other words, each variant would be calculated once (when needed), then saved to something such as a `ConcurrentHashMap`, which would be checked later.

If you're willing to try your hand at implementing the feature I mentioned, I can take another look at it. Sorry! 😅

vectorwing (changes requested):
Thanks for the contribution!

However, the way this is done on your PR might have significant performance problems. VoxelShapes should not be computed on the fly like this, as joining operations are expensive. It's sort of why I didn't do it originally. 🙁

I asked for some pointers in regards to this on Discord, and it seems this could be executed if these shapes were lazily computed and cached as they're called. In other words, each variant would be calculated once (when needed), then saved to something such as a `ConcurrentHashMap`, which would be checked later.

If you're willing to try your hand at implementing the feature I mentioned, I can take another look at it. Sorry! 😅

vectorwing:
Closing this PR for now, due to the lack of updates.
Feel free to update it with the requested changes, if you're interested, and I can look at it again. 👍

PR #980, opened 2024-08-27
A Few If-Else Function fixes

Supyovalk:


vectorwing:
Good finds! Thank you for the contribution!

PR #984, opened 2024-09-09
Update ja_jp.json

konumatakaki:
Added plant descriptions


vectorwing:
Thanks for the update!

PR #987, opened 2024-09-19
Converted block and item suppliers to DeferredBlock and DeferredItem

Lance5057:
This pr changes all Supplier<Item> to DeferredItem<Item> and all Supplier<Block> to DeferredBlock<Block>. This makes access easier with things like ItemLike.


vectorwing:
After asking around in regards to this change, it seems that it'd be better if those were changed into `Holder<>` for future-proofing, given potential updates to NeoForge.

However, updating that also means a *lot* of name changes everywhere (`.get()` into `.value()`), so I might leave it for a future release, rather than now. Thanks for the PR either way!

RaymondBlaze:
Also note that for multiloader addons, a non-platform-specific type is needed, whether it's `Holder` or `Supplier`.

PR #988, opened 2024-09-29
Update es_ar.json

axperty:
Updated many translations and added missing ones, like the skillet item, with commonly used words in Argentina, which are easier to understand for other players.


vectorwing:
Thanks for the update!

Your lang seems to contain a lot of EMI-related lines on it, as your branch was forked from FD Refabricated. FD doesn't currently have an EMI plugin in NeoForge, and I'm still analyzing whether to have one.

Could you remove these lines for now? They can be added in the future, if the plugin is implemented.

axperty:
Of course, I'll work on the requested changes, thank you for letting me know.

axperty:
Done!

vectorwing:
Nice, thank you!

vectorwing:
I'll port these changes over to 1.21, as IIRC there were no new langs there.

PR #996, opened 2024-10-04
Cutting Board Patch - 1.20

Lance5057:
#974 for 1.20


vectorwing:
So, after testing a bit, it turns out that this bug does not occur on 1.20. I had previously asked @Lance5057 to open a fix on 1.20 as well, but didn't realize it only applied to 1.21. Sorry! 😅

Closing this as there isn't anything to fix here.

PR #1008, opened 2024-11-05
Update es_mx.json

cerealconjugo:
- added 2 missed strings.
- fixed some strings here and there.
- More uncapitalized strings.


vectorwing:
Thank you for the update! ^^

PR #1009, opened 2024-11-07
Update ja_jp.json

Abbage230:
- Added missed strings
- Fixed some strings
- Update strings to match vanilla version


vectorwing:
Thanks for the update!

PR #1010, opened 2024-11-09
[1.20] Added en_nz.json

XlosVSM:
Added in translations for English (New Zealand) users. I am a Kiwi so I know what terms are used.


vectorwing:
Thanks for the new lang!

PR #1011, opened 2024-11-12
Update zh_cn.json

NumberSir:


vectorwing:
Thanks for the update! 👍

Boshigiri:
Perhaps the translation of cabinet should be unified, rather than having a mix of the both 厨柜 and 橱柜

Boshigiri:
Recommend to use 橱柜

NumberSir:
> Perhaps the translation of cabinet should be unified, rather than having a mix of the both 厨柜 and 橱柜

你重新提一个 pr 就行，这个已经合并了。

如果可以的话给 [refabricated](https://github.com/MehVahdJukaar/FarmersDelightRefabricated/blob/fabric/1.21/src/main/resources/assets/farmersdelight/lang/zh_cn.json) 也提一个。

PR #1024, opened 2024-12-02
Update ja_jp.json

Abbage230:
After reviewing the recent changes to translation approvals made by the Vanilla Crowdin proofreader and considering consistency with addon Mods and other Mods' block names, we have determined that reverting certain translations to their previous state is the most appropriate course of action.

Personally, "○○入りの木箱" feels more natural to me, but from the vanilla stage, the criteria for differentiating between "○○入り××" (水入り瓶/溶岩入りバケツ/苗木入り植木鉢) and "○○入りの××" (ハチミツ入りの瓶/溶岩入りの大釜, more common in relatively recent translations) are not unified and are so ambiguous that, to be honest, I don't think I'll ever resolve this within myself...


vectorwing:
Sounds good! Just in time for the 1.2.6 rollout, too.
I'll backport these reverts to 1.20.1 as well.

PR #1026, opened 2024-12-04
viossa translation

SquareScreamYT:


vectorwing:
Thanks for the new lang!

PR #1038, opened 2024-12-25
Update es_cl.json

Ganbare-Lucifer:
Up to latest version


vectorwing:
Thanks for the update!

PR #1041, opened 2024-12-29
Update zh_cn.json

Boshigiri:


vectorwing:
Thank you for updating pantry to cabinet! 👍

PR #1044, opened 2025-01-11
fixed a null error in CookingPotRecipeSerializer codec

ianm1647:
replacing that null with one of the existing recipe book tabs fixed an error that was being caused by trying to create a kubejs plugin for FD


vectorwing:
May I ask what kind of crash it causes, exactly? This field is designed to be nullable, as the Cooking Pot's recipe book defaults recipes without `recipe_book_tab` to the `misc` category already.

ianm1647:
see i thought the field would be fine being nulled too, but it must break something somewhere. luckily it doesn't cause a crash or anything, just a datapack error. i've tried this same thing with the build i pr'd and it worked perfectly fine. probably something on the kubejs side of things.

vectorwing:
If you still have it, please post the log with the datapack error. I might be able to check what's happening, or if I have to change something with the syntax.

ianm1647:
alright here's the latest log

https://pastebin.com/RhHXuxZY

ChiefArug:
I have also had problems with `xmap(orElse(null))` in an unrelated codec, its just something they don't seem to support.
If you want to keep the behaviour of not having to specify a tab, and not having a default, you need to make the constructor take the optional directly as a parameter and use orElse(null) in the constructor.

PR #1045, opened 2025-01-11
Cooking pot assemble fix

Lance5057:
A simple fix that makes the cooking pot use Recipe.assemble to get the result, allowing derivative recipes to edit the output item.


vectorwing:
Looks good. Do you think you could point this fix to 1.20 instead, so I can port it forward?
If not, it's alright, I can backport those lines later.

Lance5057:
Sure

PR #1058, opened 2025-02-08
Support Blueprint's `forge:buckets/water` tag

thelegitdolt:
`forge:buckets/water` is a tag added by [Blueprint](http://modrinth.com/mod/blueprint) that contains all water buckets. This is tag is used by Caverns and Chasms's golden buckets in order to determine some behavior. I add the water bucket item to the tag so there will be no difference for players not using the mod!

This change should make stoves able to be extinguished with golden water buckets, as well as making dough and cooking pot craftable using them.


vectorwing:
Looks like a reasonable unofficial tag to support.

However, I did recently change the Wheat Dough water recipe into a custom one, and I have yet to test how a tagged input would function here. I can do this on my own afterwards, though.

Could you update this PR to remove the tag addition on the Wheat Dough recipe? The Cooking Pot recipe can remain.

thelegitdolt:
all the changes are in #1126

PR #911, opened 2024-05-02
Skillet try 2

MehVahdJukaar:
Same as other PR but no junk files and for right branch

PR #1061, opened 2025-02-13
[1.20.1] Implement Refabricated Skillet tweaks.

ChrysanthCow:
This Pull Request contains two major changes to the skillet item.

If you would like to request postponement (1.3 probably?) or removal of Skillet Flipping and just have the progress bar, I'd be happy to make a new PR.

## Swap Skillets to a BEWLR
The previous solution of creating a load of models was overengineered. It'd be best to create a BEWLR and handle skillet rendering through this, this PR just brings this change from 1.21.1 over here.

## Skillet Progress Bar
The Skillet now displays a progress bar instead of its durability whilst cooking.

This is a relatively light change, all things considered, and should not break anything.

java_j9pA2Z8HJK

## Skillet Flipping
Skillet Flipping has been reworked heavily ever since #911.
The Skillet now also moves, the flipped food is only flipped 180 degrees rather than 360 degrees, and the movements are faster and less drastic in comparison to before.

You may activate this by pressing the `Attack` key whilst cooking with a skillet.

This is a breaking change on the user's end, as it adds a network handler so a packet may be sent to activate the skillet flip.

## In first person:
java_cg4WaBMqYs

## In third person:
java_NsqhniEsKX


ChrysanthCow:
This PR is now considered ready.

vectorwing:
Tested everything on both client and server with 2 players, reviewed the code, everything here looks good. 👍 

I wasn't sure whether to accept the skillet flipping mechanic, as it had to introduce networking just to exist; but easter eggs like these are the life of a mod, and I'm glad you folks took the time to implement it!

I also like the cooking progress meter. It makes handheld cooking a lot better, without guesswork.

Finally, **thank you** for freeing me from the accursed SkilletModel... With every version that arrived, this thing got more and more broken as old implementations stopped working, and I couldn't figure out how to fix it. You managed to get BEWLR posing just right, too; I'll refer to that often moving forward.

Thank you for the PR! 👏

ChrysanthCow:
Thank @MehVahdJukaar for the BEWLR posing.

You will have to rewrite for 26.1+, but luckily BEWLRs have been replaced by an entirely vanilla system.
You're always free to borrow from Farmer's Delight Refabricated if you're having trouble porting. We'll probably have done the work already.

PR #1067, opened 2025-02-25
Fixed `Nourishment` & `Comfort` effects descriptions for each lang file

mpustovoi:
Fixed the format of the key string to the correct `effect.<mod_id>.<effect_id>.description` for each available language, as the current one doesn't work by default:
image.
The same goes for the `Comfort` effect.
Test setup:
image

Updated `ru_ru.json`.

### DIFF
--- a/src/main/resources/assets/farmersdelight/lang/be_by.json
+++ b/src/main/resources/assets/farmersdelight/lang/be_by.json
@@ -205,8 +205,8 @@
   "effect.farmersdelight.nourishment": "Сытасць",
   "effect.farmersdelight.comfort": "Камфорт",
 
-  "effect.farmersdelight.nourishment.0.desc": "Папярэджвае голад, за выключэннем выкарыстання насычанасці для лячэння.",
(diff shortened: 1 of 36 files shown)


vectorwing:
Hmm, this seems like a different mod than the one these langs were supporting. The original targets used `0` in the lang to indicate the amplifier, which in the case of Comfort and Nourishment, only ever had one level.

I may need to check again to see if all effect description mods have changed to use the format you updated to. If not, it may be better to keep **both forms** of this lang, just in case of backwards compatibility.

I'll look around and see if the old one is still in use in the meantime.

mpustovoi:
> If not, it may be better to keep **both forms** of this lang, just in case of backwards compatibility.
> 
> I'll look around and see if the old one is still in use in the meantime.

I just wasn't familiar with this format. Never saw that in use.

Although, I agree with you that best way is to keep both.

You may do it yourself then since this PR was made quite a while ago and lang files could already be outdated, so feel free to close it.

vectorwing:
Found it! These langs were originally merged by a contributor, for the mod **Just Enough Effects**.

The mod shows descriptions per amplifier level. However, the mod has long since become outdated, and the latest version is for 1.16.5, so this PR is still valid to support the current effect description mods.

PR #1086, opened 2025-03-14
Wheat to Straw Cutting Recipe

ChronosVoyager:
Since straw can also be made from wheat in the real world, it feels only fitting that it's doable inside Minecraft, too :)


ChronosVoyager:
I left an extra comma in there, whoops

PR #1087, opened 2025-03-14
Wheat to Straw

ChronosVoyager:


vectorwing:
I appreciate the suggestion, but the `generated` folder isn't meant to receive manually-written data files. You would have to update FD's datagen to include it, and then `runData` in an IDE to properly update it.

Wheat is a mixed bag currently, as it has no post-processing item that would allow this recipe to make more sense (flour, semolina, wheat berries etc). I'd rather keep it as a harvest-only trait for the time being. Thanks anyway!

PR #1089, opened 2025-03-14
Make Rich Soil use a HashMap for Mushroom Colonies

Ultrasquid9:
This allows addons to add their own Mushroom Colonies without needing to mixin `RichSoilBlock`. 

This also changes the `MushroomColonyBlock` to use an `Item` instead of a `Holder<Item>`, as the `builtInRegistryHolder` method is deprecated and using an `Item` works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ out
 *.iml
 .idea
 
+# vscode
+.vscode/
+
 # gradle
 build
 .gradle
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,6 @@ out
 *.iml
 .idea
 
-# vscode
-.vscode/
-
 # gradle
 build
 .gradle
--- a/src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
@@ -47,7 +47,7 @@ public class MushroomColonyBlock extends BushBlock implements BonemealableBlock
 		.apply(builder, MushroomColonyBlock::new)
 	);
 
-	public static HashMap<ItemLike, MushroomColonyBlock> COLONIES = new HashMap<>();
+	private static HashMap<ItemLike, Block> COLONIES = new HashMap<>();
 	public static final int PLACING_LIGHT_LEVEL = 13;
 	public final ItemLike mushroomType;
 
@@ -125,6 +125,10 @@ public class MushroomColonyBlock extends BushBlock implements BonemealableBlock
 			level.setBlock(pos, state.setValue(COLONY_AGE, age + 1), 2);
 			CommonHooks.fireCropGrowPost(level, pos, state);
 		}
+	}
+
+	public static HashMap<ItemLike, Block> getMushroomColonies() {
+		return COLONIES;
 	}
 
 	@Override
--- a/src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
@@ -6,7 +6,6 @@ import com.mojang.serialization.MapCodec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 
 import net.minecraft.core.BlockPos;
-import net.minecraft.core.Holder;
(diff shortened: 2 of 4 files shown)


thelegitdolt:
i suport this

src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
```diff
@@ -34,16 +36,16 @@
...
 	public static final MapCodec<MushroomColonyBlock> CODEC = RecordCodecBuilder.mapCodec(
-			builder -> builder.group(BuiltInRegistries.ITEM.holderByNameCodec().fieldOf("mushroom").forGetter(block -> block.mushroomType), propertiesCodec())
+			builder -> builder.group(BuiltInRegistries.ITEM.byNameCodec().fieldOf("mushroom").forGetter(block -> block.mushroomType), propertiesCodec())
 					.apply(builder, MushroomColonyBlock::new)
 	);
 
+	public static HashMap<Item, MushroomColonyBlock> COLONIES = new HashMap<>();
 	public static final int PLACING_LIGHT_LEVEL = 13;
-	public final Holder<Item> mushroomType;
+	public final Item mushroomType;
```
RaymondBlaze: This is not a proper change, `Holder` should be used as addons could register mushroom colonies for their own mushrooms, where the item is registered after the blocks and thus not available in the block's constructor.

Also, the `COLONIES` field should not be exposed, use proper `register` and `get` with checks for duplicate registers, also since the `Item` is not available in the constructor, do not register to the `COLONIES` in the constructor, move it `FMLCommonSetupEvent` instead.

Since we are moving away from hardcoding, let's not restrict the value class to `MushroomColonyBlock`, and the key class could just use `Block` instead of `Item` since we are not doing the register in `MushroomColonyBlock`'s constructor anyways.

RaymondBlaze (changes requested):

.gitignore
```diff
@@ -13,6 +13,9 @@ out
 *.iml
 .idea
 
+# vscode
```
vectorwing: Not sure what occurred in your workspace, but I'd rather leave the .gitignore unchanged. Can you please revert this edit?

vectorwing (changes requested):

.gitignore
Ultrasquid9: Alright, done 👍️

src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
```diff
@@ -53,10 +59,11 @@ public class MushroomColonyBlock extends BushBlock implements BonemealableBlock
 			Block.box(1.0D, 0.0D, 1.0D, 15.0D, 14.0D, 15.0D),
 	};
 
-	public MushroomColonyBlock(Holder<Item> mushroomType, Properties properties) {
+	public MushroomColonyBlock(ItemLike mushroomType, Properties properties) {
```
RaymondBlaze, line 62: I don't think it's good to use `ItemLike` as the key, the `Item`, `Block`, `DefferedItem` and `DefferedBlock` all matches, but neither of them are equal to others by implementation.

Since you are using the `Item` to get value from the map, a modded colony which passed in a `DeferredItem` (or `DefferedBlock`) as the `mushroomType` parameter will simply fail when checked by the Rich Soil.

Please just revert this to use the `Holder<Item>` as the parameter type and key type, Holders are basically singleton and has proper `equals` and `hashCode` implementation on NeoForge, even on Fabric the Holders should still work totally fine as long as it's the `Holder.Reference` from registry.

RaymondBlaze (changes requested):

src/main/java/vectorwing/farmersdelight/common/block/MushroomColonyBlock.java
RaymondBlaze: Alternatively just move the registration away from the constructor as requested above (the Apr 7 one), which would eliminate the need of laziness of the key.

vectorwing:
Apologies for taking this long to check the PR. My bad. 😓

At first I assumed this was an implementation for 1.20.1, which would be handy. But 1.21.1 NeoForge introduces the datamap system, which makes things like these easier and datapack-oriented. So I'd rather do it this way later down.

Thanks for the PR regardless!

PR #1094, opened 2025-03-31
Fix broken Create recipes integration

mpustovoi:
Changed the `item` key in `result` to `id` as the 1.21 format introduced.


stlouisn:
fluids also need to be updated:

filling/[chocolate_pie.json](https://github.com/user-attachments/files/19654023/chocolate_pie.json)

mixing/[pie_crust_from_mixing.json](https://github.com/user-attachments/files/19654025/pie_crust_from_mixing.json)

and a couple of forge tags to change in:

mixing/[tomato_sauce_from_mixing.json](https://github.com/user-attachments/files/19654093/tomato_sauce_from_mixing.json)

mpustovoi:
Do you mean replacing the namespace `forge` with `neoforge` in these recipes?

I haven't seen any errors with these ones in the log, so I don't quite get it.

mpustovoi:
Oh, that's it, I got that these are files, not links.

stlouisn:
Yes just those three recipes need further changes for 1.21.1

src/main/resources/data/farmersdelight/recipe/integration/create/filling/chocolate_pie.json
```diff
@@ -11,13 +11,14 @@
       "item": "farmersdelight:pie_crust"
     },
     {
-      "fluidTag": "forge:chocolate",
-      "amount": 500
+      "type": "fluid_stack",
+      "amount": 500,
+      "fluid": "create:chocolate"
```
RaymondBlaze: Please use the `c:chocolates` tag instead.

src/main/resources/data/farmersdelight/recipe/integration/create/mixing/pie_crust_from_mixing.json
```diff
@@ -17,13 +17,14 @@
       "item": "create:wheat_flour"
     },
     {
-      "fluidTag": "forge:milk",
-      "amount": 250
+      "type": "fluid_stack",
+      "amount": 250,
+      "fluid": "minecraft:milk"
```
RaymondBlaze: Please use the `c:milk` tag instead.

RaymondBlaze (changes requested):

src/main/resources/data/farmersdelight/recipe/integration/create/filling/chocolate_pie.json
```diff
@@ -11,13 +11,14 @@
       "item": "farmersdelight:pie_crust"
     },
     {
-      "fluidTag": "forge:chocolate",
-      "amount": 500
+      "type": "fluid_tag",
+      "amount": 500,
+      "fluid_tag": "c:chocolate"
```
RaymondBlaze: For some reason Create decided to use the plural form `c:chocolates`, which does makes sense if chocolate variants exists, like `c:chocolates/black`. However Create didn't decide to add a subtag for its plain (or milk) chocolate.

Anyways, for the recipe to work at least for Create's chocolate and other mods that opt-in that naming, please use `c:chocolates` here, if it feels wrong to support Black Chocolate and White Chocolate from other mods, revert to just using the Create's `create: chocolate`.

RaymondBlaze (changes requested):

src/main/resources/data/farmersdelight/recipe/integration/create/filling/chocolate_pie.json
mpustovoi: I guess since this folder is dedicated to Create, then there should only be something related to Crete.

RaymondBlaze (approved):

SirRosenfeld (approved):
Independent analysis on my side came to the same solution. Can confirm this fixed it for me too before I saw someone already submitted those changes.

vectorwing:
Looks good to me; the syntax matches the new recipe formats.
Thank you for taking the time to fix these integration recipes!

PR #1109, opened 2025-04-25
Fix recipe IDs not showing in JEI

Abbie5:


vectorwing:
Is this a new change from JEI? I don't recall this bug happening before.
If it is, I can merge it right away, just wanted to confirm.

RaymondBlaze:
> Is this a new change from JEI? I don't recall this bug happening before. If it is, I can merge it right away, just wanted to confirm.

It's due to the vanilla change moving from holding recipe ids in the `Recipe` itself to use `RecipeHolder` instead, which results in JEI also change to use the `RecipeHolder` for the categories generic type so recipe id may be retrieved.

Abbie5:
This change was made in JEI in 1.20.2, which is why the bug wasn't present in the 1.20.1 version of Farmer's Delight.

vectorwing:
It took me a while to realize where the recipe ID would show up, but I found it:
image

Yes, this PR fixes it not being shown with debug info. I'll merge it in a moment. Thank you!

PR #1112, opened 2025-05-01
Fix rich soil farmland turning into dirt and negating fall damage

SarahIsWeird:
This is a twofer:
1. `RichSoilFarmlandBlock::fallOn` has to call `Entity::causeFallDamage`, otherwise if you fall onto it, you take no fall damage.
2. There are certain interactions with other mods that cause the rich soil farmland to turn into dirt. I noticed this with Allurement's Shockwave enchantment, which turns surrounding farm land back into dirt by calling `FarmBlock::turnToDirt`. The mixin just checks at the top of that static function if it's a `RichSoilFarmlandBlock` and returns early if so. Some friends and me did some brainstorming and weren't able to come up with a situation where this would cause an issue.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/RichSoilFarmlandBlock.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/RichSoilFarmlandBlock.java
@@ -104,7 +104,8 @@ public class RichSoilFarmlandBlock extends FarmBlock
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


vectorwing:
Good fixes! 👍

I tested using just the fall damage fix with Allurement, and it didn't seem to affect Rich Soil whatsoever, so I'm unsure how necessary the mixin is, since it just seems to silently fail for Rich Soil Farmland. I'll take a look later.

PR #1126, opened 2025-05-19
Support blueprint's `forge:buckets/water` tag

thelegitdolt:
Hi vector i've no idea how git works and i think i broke something whlie trying to sync the fork, but in this PR i've did all the requested changes (deleted the dough recipe)


thelegitdolt:
I tested myself, tagged input also works fine with vanilla water buckets. If you would rather do it yourself, i can revert the latest commit

thelegitdolt:
https://github.com/user-attachments/assets/98df4173-8f18-473a-b1b3-3004f78819a5

also works with modded buckets

src/main/java/vectorwing/farmersdelight/common/block/StoveBlock.java
```diff
@@ -63,13 +64,13 @@ public InteractionResult use(BlockState state, Level level, BlockPos pos, Player
 				extinguish(state, level, pos);
 				heldStack.hurtAndBreak(1, player, action -> action.broadcastBreakEvent(hand));
 				return InteractionResult.SUCCESS;
-			} else if (heldItem == Items.WATER_BUCKET) {
+			} else if (heldItem.builtInRegistryHolder().is(ForgeTags.BUCKETS_WATER)) {
```
vectorwing: I think `builtInRegistryHolder()` is marked for deprecation. You can use `heldStack.is(ForgeTags.BUCKETS_WATER)` here instead, a field from a few lines above.

vectorwing (changes requested):

src/main/java/vectorwing/farmersdelight/common/block/StoveBlock.java
thelegitdolt: oh yeah

vectorwing: Aside from this, the PR looks fine. I can merge after it.

PR #1129, opened 2025-05-31
[1.21] move worldgen and damage types to datagen

Abbie5:


vectorwing:
Hi! Sorry for the long delay in my reply.

Would these changes also be compatible with 1.20 without much change, or do they rely on 1.21 features?

Abbie5:
Hi, there are not any 1.21 specific features used in this PR, so it's easy to port to the 1.20 branch. I have done so in #1242.

vectorwing:
I merged the 1.20 version of this PR on `dev/1.3`. Since both PR's seem to be mostly identical, and nothing differs for 1.21 syntax, I'll close this one, and port over the changes from 1.20 at a later date.

Thank you again for the PR! 👍

PR #1130, opened 2025-06-02
Updated italian localisation [25-06-02]

BlackShadow77:
Updated Italian localisation to fit latest updates


src/main/resources/assets/farmersdelight/lang/it_it.json
```diff
@@ -192,7 +213,7 @@
 
   "itemGroup.farmersdelight": "Farmer's Delight",
 
-  "farmersdelight.container.recipe_book.cookable": "Mostra ciò che si può cuocere",
+  "farmersdelight.container.recipe_book.cookable": "Mostra cio' che si puo' cuocere",
```
vectorwing: Quick question: why are instances of reverse-accents being converted to a loose quote mark?
Although I don't speak Italian, this seems like the wrong form of writing this accent. Just wanted to confirm.

src/main/resources/assets/farmersdelight/lang/it_it.json
```diff
@@ -48,6 +51,24 @@
   "block.farmersdelight.green_canvas_sign": "Cartello di tela verde",
   "block.farmersdelight.red_canvas_sign": "Cartello di tela rosso",
   "block.farmersdelight.black_canvas_sign": "Cartello di tela nero",
+
+  "block.farmersdelight.hanging_canvas_sign": "Cartello di tela pensile",
+  "block.farmersdelight.white_hanging_canvas_sign": "artello di tela bianco pensile",
```
VoidLeech: Missing a 'C' here

src/main/resources/assets/farmersdelight/lang/it_it.json
vectorwing: Yeah, the base font can support accented letters already, so this change isn't necessary.
image
Could you please revert the accent changes?

vectorwing (changes requested):

src/main/resources/assets/farmersdelight/lang/it_it.json
BlackShadow77: This is a very nice question that show you put brain and heart in the project and i like it.

The answer is pretty simple: _**character encoding**_.

The accents usually create tons of problems in storing/showing on IT systems so, to keep things simple, we use `'` char to identify an accented letter. I think in english it's called "[typographic approximation](https://en.wikipedia.org/wiki/Typographic_approximation)".

In Italy we also have an institution ([Accademia della Crusca](https://en.wikipedia.org/wiki/Accademia_della_Crusca)). It serve as guide to italian language rules. I also can link an article (unfortunately there is no traslation) about the use of `'` instead accents [there](https://accademiadellacrusca.it/it/consulenza/uso-del-digramma-letteraapice-in-sostituzione-della-lettera-accentata/89).

In conclusion: is only a (old) convention about writing some characters in IT systems. If you prefer i can change back to old style.

src/main/resources/assets/farmersdelight/lang/it_it.json
BlackShadow77: Ty, i'll correct that

src/main/resources/assets/farmersdelight/lang/it_it.json
BlackShadow77: As you wish.
I'll work on that

vectorwing: Ahh, that makes a lot of sense! Thanks for explaining.

I'm a native Portuguese speaker, so I remembered when older MC versions mixed the Unicode font with the standard one to show accented letters, which used to look horrible. I assumed the change was due to this issue at first.

As I mentioned in the other comment, it does seem like Minecraft can handle accented letters at least, so it may be worthwhile keeping them on 1.20+.

BlackShadow77: No problem, it's fine for me explain what i done because i'm always open to new prospectives and ready to correct my errors, eventually.

As i said for italians is pretty the same, to be honest i needed **CTRL+F** to find all occurence because i can read that as accent with fluency.

I converted that as a draft and already done the changes as requested: your wishes are my duty!

I want to check again the file later before include it in final version to be sure nothing is left. Also noticed some old translations need some love so i'll also work on that meanwhile.

I'll wait for a portuguese lesson sooner or later

vectorwing (approved):

vectorwing:
Thanks for the updates! Sorry for the delay in approving them. 😅

BlackShadow77:
No problem, sorry if i didn't mark as ready before. Busy with life stuffs

PR #1153, opened 2025-07-06
Add lang hu_hu

bayi:
Added Hungarian translation.


vectorwing:
Thanks for the new translation!

PR #1159, opened 2025-07-24
Make the CookingPotRecipeBookTab enum StringRepresentable

QinomeD:
I noticed that the `CookingPotRecipeBookTab` enum wasn't `StringRepresentable` when updating KubeJS Delight to 1.21.1, when I tried to add it as a enum recipe component, which requires the enum to be `StringRepresentable`.

This change would make it so I don't need to work around with a duplicate of the enum that implements that interface, and of course the addon would adapt to any future changes (idk if there would be any, but just in case)


vectorwing:
Makes sense. I seem to already use it in another enum I created.
I'll merge this PR and backport it to 1.20 as well.

QinomeD:
Splendid, thanks!

PR #1163, opened 2025-08-06
[1.21] Make wild cabbage and beetroot compatible with any beach biome

lyaneii:
Wild cabbages and wild beetroots still seem to use some old behaviour, but the main change here is this hardcoded `minecraft:beach` for the allowed biome, changing it to `#minecraft:is_beach` will allow it to be compatible with other world generation mods where no regular `minecraft:beach` might exist. *(speaking about Still Life Mod/Datapack, I had this issue there :c)*
Here is a before/after for demonstration, they are located around the same area with the same type of custom beach generated by Still Life.
Before:
screenshot of a custom beach biome generation by still life with wild cabbages and wild beetroots not generating
After:
screenshot of a custom beach biome generation by still life with the change of making wild cabbages and wild beetroots compatible with it

### DIFF
--- a/src/main/resources/data/farmersdelight/neoforge/biome_modifier/wild_beetroots.json
+++ b/src/main/resources/data/farmersdelight/neoforge/biome_modifier/wild_beetroots.json
@@ -1,6 +1,6 @@
 {
-  "type": "neoforge:add_features",
-  "biomes": "minecraft:beach",
+  "type": "farmersdelight:add_features_by_filter",
+  "allowed_biomes": "#minecraft:is_beach",
   "features": "farmersdelight:patch_wild_beetroots",
   "step": "vegetal_decoration"
 }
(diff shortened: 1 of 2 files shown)


vectorwing:
I was trying to recall if there was a reason for me to hardcode them to specifically the vanilla beach biome, since this tag exists on 1.20.1 as well. And the only thing I can think of was avoiding Snowy Beach, which is snow-coated.

This beach exists in the `is_beach` tag, and I feared it would look odd to have these plants thriving in climate cold enough to snow. But it doesn't look too bad, I guess!

image

vectorwing:
I guess there is no harm in letting other beach biomes spawn it. I can always introduce a deny tag later, if necessary.

@lyaneii, could you make this PR for the `1.20` branch as well? I can add it to the current 1.2.9 patch on both versions.

lyaneii:
Thank you, created PR #1173 for 1.20 as well.
Regarding the climate thing, maybe the `"min_temperature"` and `"max_temperature"` can be added to them as well, they are currently not configured for these two crops.
For any other awkward beaches I could think of, there are maybe stony beaches or gravel beaches, but I don't think standard minecraft introduces those, or at least not under the `#minecraft:is_beach` tag. But also not sure if this setting can be driven by a datapack either, would open up more customisation for modded biomes next to the default values.

vectorwing:
I believe these patches will fail anyway on stony shores, as they require some form of `farmersdelight:terrain` to be placed on.

Thank you for both PR's!

PR #1173, opened 2025-08-21
[1.20] Make wild cabbage and beetroot compatible with any beach biome

lyaneii:
1.20 version of this PR #1163


vectorwing:
Nice! Thank you for making this one too! 👍

PR #1175, opened 2025-08-22
Update CuttingBoardRecipeBuilder.java

brnbrd:
The private constructor for CuttingBoardRecipeBuilder accepts `chance` as type float, however this `cuttingRecipe` method accepts and casts it to type int instead. This means that the current RecipeBuilder can only build recipes with 100% chance for the main result. This commit fixes the type of `chance` to be float in the builder method.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/data/builder/CuttingBoardRecipeBuilder.java
+++ b/src/main/java/vectorwing/farmersdelight/data/builder/CuttingBoardRecipeBuilder.java
@@ -45,7 +45,7 @@ public class CuttingBoardRecipeBuilder
 	/**
 	 * Creates a new builder for a cutting recipe, providing a chance for the main output to drop.
 	 */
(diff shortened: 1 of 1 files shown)


vectorwing:
I was confused for a moment, since I do use that builder to generate recipes with float chances... until I realized that the method you updated is unused in FD's datagen. 😅

Generally, FD doesn't make the primary output have a chance, only assigning it to secondary ones, using the `addResultWithChance()` method. So I never caught that this builder method had `chance` as an int.

Thanks for the fix. Are you and add-on creators using that method for chances on primary recipes? Just checking whether to obsolete the method or keep it around.

brnbrd:
> Are you and add-on creators using that method for chances on primary recipes

I am not currently using the method, but was considering it, which is what led me to this issue. I've seen chance primary results in some other add-ons, so I would argue the method has a place to stay for some niche cases.

I am also now realizing I made this exact PR before (#929), and that change got merged to the dev/1.3 branch but not to 1.20.

PR #1181, opened 2025-08-30
Update Hungarian Translations

bayi:
Translated the new keys


vectorwing:
Thank you for the update! 👍

PR #1184, opened 2025-09-02
Add a cabinets block tag, and a Mushroom Colonies item and block tag

thelegitdolt:
I refactored the path of the current`ModTags#CABINETS` to `ModTags#CABINET_ITEMS` and `ModTags#WOODEN_CABINETS` to `ModTags#WOODEN_CABINET_ITEMS`, following the convention set by the wild crop tags. This might break the code of addons running datagen, but shouldn't cause a crash.


thelegitdolt:
Closes https://github.com/vectorwing/FarmersDelight/discussions/902

thelegitdolt:
Since I've the fork and all, I've added a `#farmersdelight:mushroom_colonies` tag which contains the mushroom colonies. I've also made `#farmersdelight:unaffected_by_rich_soil` use this tag instead of manually listing the mushroom colonies

vectorwing:
Hi! Sorry for the long delay.

These tags seem reasonable. Could you point the PR towards the `dev/1.3` branch? There may be a few conflicts in ItemTags, but not much else has changed. @thelegitdolt

vectorwing:
Thanks for pointing it to `dev/1.3`! Everything here looks good. 👍

PR #1192, opened 2025-09-12
Allow villagers to eat simple vegetables

isErenG:
Villagers weren't able to consume some of the items they picked up (cabbage, tomatoe,  rice, onion). Created a simple mixin to modify the `FOOD_POINTS` map.

I didn't give much thought to the food point worth so feel free to change it accordingly. I think it is fine though.


KreloX:
It's not up to me to pull this or not but Farmer's Delight uses the following code along with accesstransformers to modify hard coded collections, so this implementation wouldn't be consistent with the rest of the mod.
https://github.com/vectorwing/FarmersDelight/blob/afc79de755ae04d20ed9654ab4056e53a717cf32/src/main/java/vectorwing/farmersdelight/common/CommonSetup.java#L119-L142

isErenG:
I saw this file but the `FOOD_POINTS` map is immutable in `Villager.class`. That's why I resorted to making a mixin instead.

KreloX:
the private and final modifiers can be changed using the accesstransformer file
https://github.com/vectorwing/FarmersDelight/blob/afc79de755ae04d20ed9654ab4056e53a717cf32/src/main/resources/META-INF/accesstransformer.cfg#L5-L8

isErenG:
Oh I didn't know, I'll rewrite it then. Tysm

vectorwing:
Tested the PR just now, and it works like a charm! It really was that simple, huh...
Thank you for the change! This basically concludes the integration of villagers with FD's crops. ❤️

PR #1195, opened 2025-09-21
Exclude .cache upon compilation

JerryLu086:
Should work with 1.21.1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -139,6 +139,10 @@ dependencies {
 
 }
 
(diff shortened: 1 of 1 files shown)


vectorwing:
I guess it's worth not adding it to the final .jar indeed. Though after consulting with other devs, this line may be better added to something like `sourceSets`.

I'll merge and move it there manually. Thank you for the PR! 👍

PR #1196, opened 2025-10-02
QuickMoving food from Cooking pot output to player inventory gives crafting xp

medi-torimorta:
call Slot#onQuickCraft() in ,QuickMoveStack, much like abstractFurnaceMenu does.
quickMoving food from the result slot into player's inventory should now give XP / achievement, just like when they would've if they dragged the item.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/entity/container/CookingPotMenu.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/entity/container/CookingPotMenu.java
@@ -121,6 +121,7 @@ public class CookingPotMenu extends RecipeBookMenu<RecipeWrapper>
 				if (!this.moveItemStackTo(slotStack, startPlayerInv, endPlayerInv, true)) {
 					return ItemStack.EMPTY;
 				}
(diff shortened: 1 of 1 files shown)


vectorwing:
After testing a few times, it seems shift-clicking out of the output slot (which is what I assume "quick crafting" is) already delivers EXP to the player, as the method eventually lands on `slot.onTake()`, which handles that part.

Is there a specific interaction this PR solves? I'm not sure if I understood it correctly. @medi-torimorta

medi-torimorta:
Hi, though not completely sure what I was thinking at the time, I am guessing this PR was a result of a mix-up on my part.

I found out that cooking pots doesn't give exp when players right-clicks a pot containing cooked food with a bowl (or any other container) to directly take the food without opening the menu.
I was also doing a thing with my own mod that had to do with the quick-moving, and presumably mixed those up.

sorry for the confusion, i will close this PR.

vectorwing:
Ah, I see! Thanks for clarifying!

Yeah, EXP isn't currently given from interactions other than UI extraction, a bit in the style of furnaces. Though I can consider changing it in the future.

PR #1208, opened 2025-11-01
Update Classical Chinese

BeiDou114514:
Update the localization.


vectorwing:
Thanks for the update!

PR #1221, opened 2025-12-25
Fix the logic of the useHeldItemOnMeal method for the cooking pot and…

VBlackCAT:
… improve the handling of dishes without containers.

The following issue has been resolved:
When there is more than one stack of dishes in the pot that do not require containers, if the player right-clicks while holding the same type of dish, the dish would be incorrectly identified as a container and consumed. 
The useHeldItemOnMeal method has been modified to correctly handle the following scenarios: For dishes without containers, empty-handed interaction is allowed to take one portion. When the player right-clicks while holding the same type of dish, it will only open the pot's menu.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/block/entity/CookingPotBlockEntity.java
+++ b/src/main/java/vectorwing/farmersdelight/common/block/entity/CookingPotBlockEntity.java
@@ -468,14 +468,33 @@ public class CookingPotBlockEntity extends SyncedBlockEntity implements MenuProv
 		}
 	}
 
-	public ItemStack useHeldItemOnMeal(ItemStack container) {
-		if (isContainerValid(container) && !getMeal().isEmpty()) {
-			container.shrink(1);
-			inventoryChanged();
-			return getMeal().split(1);
-		}
-		return ItemStack.EMPTY;
-	}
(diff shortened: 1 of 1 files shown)


vectorwing:
It took me a while to realize what you meant, as the pot auto-ejects meals which don't require containers...

But I assume you mean the state pictured below, right?

image

This is an interaction which completely slipped past my radar; I forgot that the "output" slot would not block cooking, since the "meal" slot is still empty. I assumed there would never be a situation where a non-container meal would sit in that slot, but I guess I was wrong. 😆

I can take a look at the PR later and review the solution. Though in this case, I think a few extra conditions are required, such as not displaying "Served on:" when hovering the meal. I can change that later, though.

image

vectorwing:
I reviewed the bug, and decided to fix it in a different way than the PR proposes.

Turns out I had a weird final return on `isContainerValid()` which checked if the used item and the meal were the same. Dunno why I ever added that, as there is no use case for that; removing it fixes the meal deletion bug.

As for the feature of grabbing a serving if using the same meal on the pot, it feels a bit unintuitive. For non-contained items, I think just opening the pot UI and grabbing them is simpler; but thanks for proposing it!

I also fixed the empty "Served on:" line on non-contained meal items.

Either way, thank you for taking the time to report and fix the bug! I'll credit you on the changelog. ^^

PR #1222, opened 2025-12-31
Added & adjusted effectTick checks for the Comfort effect

TNWing:
Currently, the comfort effect does not respect the natural regeneration gamerule, so I added a flag for that.

Also, to account for a few situations regarding food & saturation values, I swapped out the saturation check for food level checks (although I am uncertain if the main branch's behavior is actually intended. If it is, then this change can be discarded.)

These are the scenarios that I was accounting for:
1. Hunger bar greater than or equal to 18 with no saturation
2. Hunger bar less than 18 with no saturation
3. Hunger bar less than 18 with saturation

Situation 2 works as intended, and the player is healed. The concerns are situations 1 and 3.
In situation 1, as comfort is applied independently from natural regeneration, this can result in the player healing 2 HP instead of just 1.
In situation 3, comfort doesn't trigger despite not being able to consume saturation for fast healing. For instance, if I eat a beef stew when I am at 0 hunger, even though I'm below the threshold for natural regeneration, I will not heal due to having saturation. 
By checking food levels instead of saturation, this ensures the player only heals if there is no way for natural regeneration to occur.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/effect/ComfortEffect.java
+++ b/src/main/java/vectorwing/farmersdelight/common/effect/ComfortEffect.java
@@ -6,6 +6,7 @@ import net.minecraft.world.effect.MobEffectCategory;
 import net.minecraft.world.effect.MobEffects;
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.GameRules;
 
 @SuppressWarnings("unused")
 public class ComfortEffect extends MobEffect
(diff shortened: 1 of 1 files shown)


vectorwing:
Hello! Thank you for the PR. 👍 

> Currently, the comfort effect does not respect the natural regeneration gamerule, so I added a flag for that.

The natural regeneration rule is not synced between client and server. The client has no way of knowing if it's on or off, as food state is managed by the server. We would have to manually update this gamerule via packets to make this work, which I opted not to do, due to the complexity involved, and the fear of messing things up.

Your current implementation won't have an effect on servers, so I suggest removing it. Sorry! 😓

> By checking food levels instead of saturation, this ensures the player only heals if there is no way for natural regeneration to occur.

Saturation is quirky, and there are indeed scenarios where it doesn't quite deplete completely when you're hungry enough, preventing Comfort from kicking in when it should. In such cases, the game takes a really long time to flush out the few remaining decimals before it resorts to food healing. This makes it seem like it's not working, but it will eventually, after several seconds. If it doesn't, that saturation has to be manually flushed out via exhaustion (sprinting, jumping etc) before it can work. The woes of working with MC's hunger bar, sadly...

That said, I suppose a food value check could work as well. I'll have to test this locally before I can confirm it. That said, I also have plans to rework Comfort and Nourishment into one, though this may occur down the line.

vectorwing:
Hello! Quick update about this PR:

I have recently passed a change which merges Comfort's use into Nourishment, so Comfort will be retired in the upcoming version. I'll be closing this PR, as the code will no longer be used.

Thanks for the contribution anyway! 👍

PR #1228, opened 2026-02-03
Update CuttingBoardRecipeBuilder.java

QTpyeRose:
enable public reference of constructor, allow class extensions for custom implementations to piggyback on in add-on mods.

the pot recipe builder is public. allowing code of this nature.
```java
private static class PotBuilder extends CookingPotRecipeBuilder {
        private final String name;

        public PotBuilder(ItemLike result, int count, int cookingTime, float experience, @Nullable ItemLike container) {
            super(result, count, cookingTime, experience, container);
            this.name = "cookingpot_" + BuiltInRegistries.ITEM.getKey(result.asItem()).getPath();
        }

        public static CookingPotRecipeBuilder cookingPotRecipe(ItemLike mainResult, int count, int cookingTime, float experience) {
            return new PotBuilder(mainResult, count, cookingTime, experience, (ItemLike) null);
        }

        public static CookingPotRecipeBuilder cookingPotRecipe(ItemLike mainResult, int count, int cookingTime, float experience, ItemLike container) {
            return new PotBuilder(mainResult, count, cookingTime, experience, container);
        }

        @Override
        public void build(RecipeOutput output) {
            this.save(output, ResourceLocation.fromNamespaceAndPath(Qtpyedelight.MODID, name));
        }
    }
```
this creates and extend the class for a custom implementation. enabling an add-on to use the recipe builder built into the mod for there own purposes but altered to write to there own namespace. this allows add-ons to use the recepie builder instead of reimplimenting there own one from scratch.

however the cutting recipe only has  private constructors making an extends class implementation impossible.

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/data/builder/CuttingBoardRecipeBuilder.java
+++ b/src/main/java/vectorwing/farmersdelight/data/builder/CuttingBoardRecipeBuilder.java
@@ -29,7 +29,7 @@ public class CuttingBoardRecipeBuilder implements RecipeBuilder
 	private final Ingredient tool;
 	private SoundEvent soundEvent;
 
-	private CuttingBoardRecipeBuilder(Ingredient ingredient, Ingredient tool, ItemLike mainResult, int count, float chance) {
+	public CuttingBoardRecipeBuilder(Ingredient ingredient, Ingredient tool, ItemLike mainResult, int count, float chance) {
 		this.results.add(new ChanceResult(new ItemStack(mainResult.asItem(), count), chance));
 		this.ingredient = ingredient;
 		this.tool = tool;


vectorwing:
Hello! Thanks for the PR!

This recipe builder is already changed to public on the upcoming 1.3 release of FD, on the `dev/1.3` branch. That said, maybe I could sneak this in if I make another patch release.

vectorwing:
Actually, nevermind, I mixed things up. The constructors aren't public yet.
I can update this soon, though I'm unsure when it will be released.

QTpyeRose:
also in general if you looking to add more stuff for people who develop addons, considere adding in a namespace argument into the constructor for recpie builders. and save it to a namespace variable in the builder and reference that whenever you save the items to the folder.

this allows other to directly use your recpie builders and write to there own mods folder without having to a bunch of overrides like i posted in the original code.

also consider making a food presets class that holds stuff like the common reused food builders (it cant be an enum cause each call needs to be a new instance)

something like
```java
public class FoodPresets {
    public static FoodProperties.Builder comfortShort() {
        return new FoodProperties.Builder()
                .nutrition(8).saturationModifier(0.7f)
                .effect(() -> comfort(SHORT_DURATION), 1.0F);
    }

    public static FoodProperties.Builder comfortMedium() {
        return new FoodProperties.Builder()
                .nutrition(12).saturationModifier(0.8f)
                .effect(() -> comfort(MEDIUM_DURATION), 1.0F);
    }

    public static FoodProperties.Builder comfortLong() {
        return new FoodProperties.Builder()
                .nutrition(14).saturationModifier(0.75f)
                .effect(() -> comfort(LONG_DURATION), 1.0F);
    }

    public static FoodProperties.Builder nourishShort() {
        return new FoodProperties.Builder()
                .nutrition(10).saturationModifier(0.6f)
                .effect(() -> nourishment(SHORT_DURATION), 1.0F);
    }

    public static FoodProperties.Builder nourishMedium() {
        return new FoodProperties.Builder()
                .nutrition(12).saturationModifier(0.8f)
                .effect(() -> nourishment(MEDIUM_DURATION), 1.0F);
    }

    public static FoodProperties.Builder nourishLong() {
        return new FoodProperties.Builder()
                .nutrition(14).saturationModifier(0.75f)
                .effect(() -> nourishment(LONG_DURATION), 1.0F);
    }
}
```
and then use that in your food definitions.

would enable addons to hook into your mod as a library and used those easier.
the call would be `nourishLong().build(output)` instead of having the whole segment every time.

QTpyeRose:
or you could set it to have conformtFood and nourishmentFood as jsut 2 functions and take in the time for them

vectorwing:
Alright, I'm back.

Due to the existence of an older PR which elaborates a little further onto making recipe builders subclassable, and also points to 1.20, I'll prioritize ir over this one.

Thank you for the PR regardless! I'll tweak the other PR in post to accomodate for utilities and suggestions. 👍

PR #1230, opened 2026-02-05
feat: zh_hk translations

Duppy-Conqueror:


vectorwing:
Thank you for the new lang!

PR #1235, opened 2026-02-12
Adjust fried egg recipe to allow other egg types

DarkFyre42:
Fried eggs currently can only be made by cooking the normal egg, excluding even brown or blue eggs. This fixes that.

### DIFF
--- a/src/generated/resources/data/farmersdelight/recipes/fried_egg.json
+++ b/src/generated/resources/data/farmersdelight/recipes/fried_egg.json
@@ -4,7 +4,7 @@
   "cookingtime": 200,
   "experience": 0.35,
   "ingredient": {
-    "item": "minecraft:egg"
+    "tag": "forge:eggs"
   },
   "result": "farmersdelight:fried_egg"
 }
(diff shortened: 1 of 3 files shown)


vectorwing:
This is intentional.

Mods should not add smelting recipes with tags as inputs. This can cause recipe conflicts with modded eggs, as the recipe only has a single input, and the modded egg may cook into something different. The Furnace cannot resolve these conflicts in-game.

Instead, a specific recipe must be created for each egg variant. I will do this once FD is ported to a version of Minecraft which includes the cold and warm eggs, but for now, it should remain as-is, and be incremented by modpack makers through data packs.

Thank you for the PR, either way! 👍

vectorwing:
Additionally, when proposing datapack changes, be careful which files you edit. You edited files under the `generated` path; those are automatically generated by FD's "datagen" code. Your changes would be wiped in the next `runData` we did.

Those changes should be made within FD's generator classes, which are under the `data` path in the project code.

DarkFyre42:
Ah I see, thanks for the clarification. It was a minor pain point for a player on my server since they put their base in a place without normal chickens. Easily solvable, but a but confusing for them.

Apologies for editing in the wrong place. If you can't tell I've not messed with mods before 😅

Thanks again for your reply!

PR #1240, opened 2026-02-21
Make Food Block Shapes Rotatable

TheGridExpert:
As of now, most food blocks in the mod change their model depending on how many servings remain. However, the collision shape stays the same regardless of the model, which can look off in certain situations. This PR makes the shapes of most such blocks change depending on the direction, so they appear correctly when placed and eaten. The affected blocks are: all pie variants, the roast chicken, the honey glazed ham, and the shepherd's pie. I've decided to leave the rice roll medley as it is, as its model is too detailed and this might be too much.

Shape rotation is handled mathematically in the added ShapeUtils class. Rotated shape arrays are cached, so the game doesn't need to recompute this during runtime. For plated feast blocks specifically, the computed shape arrays are also cached in ShapeUtils itself. This is useful if multiple blocks use the same shape, especially in addons, so less calculations and extra classes will be required. Thence, RoastChickenBlock, HoneyGlazedHamBlock, and ShepherdsPieBlock have been removed, as the only thing different in them was their shapes. A new RotatedFeastBlock class handles serving-dependent shapes with and without a plate. The shapes themselves were moved into a separate BlockShapes class. For now, it is placed in the common folder alongside FoodValues, but feel free to move it wherever it looks better, wasn't sure where to put it.


vectorwing:
After testing locally and checking the code and references, everything looks good so far. 👍

I'll do another rundown and post any questions I have before merge soon.

src/main/java/vectorwing/farmersdelight/common/utility/ShapeUtils.java
```diff
@@ -0,0 +1,149 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+
+/**
+ * Credits to TheGridExpert for the implementation
+ */
+public class ShapeUtils {
+
+    private static final Map<VoxelShape[], VoxelShape[][]> PLATED_SHAPE_CACHE = new IdentityHashMap<>();
```
vectorwing, line 17: Are the cached shapes inside ShapeUtils meant to be "defaults" for RotatedFeastBlock users which don't specify their own shapes?

Just making sure, as other feasts do have their own specific shapes in BlockShapes.

src/main/java/vectorwing/farmersdelight/common/BlockShapes.java
```diff
@@ -0,0 +1,29 @@
...
+            Block.box(4.0D, 2.0D, 6.0D, 12.0D, 9.0D, 12.0D),
+            Block.box(4.0D, 2.0D, 4.0D, 12.0D, 9.0D, 12.0D)
+    };
+    public static final VoxelShape[] HONEY_GLAZED_HAM_SHAPES = new VoxelShape[] {
+            Block.box(4.0D, 2.0D, 4.0D, 12.0D, 4.0D, 12.0D),
+            Block.box(4.0D, 2.0D, 8.0D, 12.0D, 10.0D, 12.0D),
+            Block.box(4.0D, 2.0D, 6.0D, 12.0D, 10.0D, 12.0D),
+            Block.box(4.0D, 2.0D, 4.0D, 12.0D, 10.0D, 12.0D)
+    };
+    public static final VoxelShape[] SHEPHERDS_PIE_BLOCK = new VoxelShape[] {
```
vectorwing: I suppose this field could be called `SHEPHERDS_PIE_SHAPES`, to be consistent with the others.

src/main/java/vectorwing/farmersdelight/common/utility/ShapeUtils.java
TheGridExpert: These are not really defaults, but rather a "newbie mistake catcher" for addons. I've seen many addons add arrays of similar foods, for example jellies, which only differ in colour. If the addon developer uses RotatedFeastBlock without considering how the rotation works, the game will end up calculating the same over and over. This additional caching is there in such cases. It does not make it worse for the base mod, but it is useful if there are a few Farmer's Delight addons in the mod pack.

src/main/java/vectorwing/farmersdelight/common/BlockShapes.java
TheGridExpert: Yup, didn't notice. Fixed.

src/main/java/vectorwing/farmersdelight/common/utility/ShapeUtils.java
vectorwing: Makes sense. Thanks for explaining!

PR #1241, opened 2026-02-22
Replace use of RegistryObject with Supplier

cassiancc:
This PR replaces the use of the Forge-specific `RegistryObject` with the more universal `Supplier`. This increases compatibility with addons wishing to support both the official Farmer's Delight on Forge and Farmer's Delight Refabricated, or for people using Farmer's Delight addons through tools like Sinytra Connector or Kilt. This also brings the 1.20 version further into parity with the 1.21 version, which already uses Suppliers.


cassiancc:
Closed to avoid breaking existing Forge addons.

PR #1242, opened 2026-02-25
[1.20] move worldgen and damage types to datagen

Abbie5:
same as #1129 but for 1.20


vectorwing:
Thanks for the PR!

I'll have to sit down and understand the changes first, as I think I've only used this form of datagen on Enchantments, during the 1.21.1 port. I'll get back with my review later.

vectorwing:
I pointed the PR to `dev/1.3` and there were no conflicts or differences. 👍

vectorwing:
After studying the code a bit more, and comparing to the other datagen PR for similarities (#959), I think I get what's going on now, and I should be able to maintain it.

I had a few questions looking over, but I think I answered them to myself in the end. Aside from code organization (which I'll do on my own after this), this PR is good for a merge.

Since the 1.21-bound PR seems to be exactly the same, I might close that one later and simply port the 1.20 changes over when it's time, depending on how my refactoring goes.

Thank you for the contribution! ^^

PR #1246, opened 2026-03-06
Fix Canvas Rug & Mats missing proper footstep sounds

Mickeon:
When stepping onto a Canvas Rug, Half Tatami Mat and Full Tatami Mat, the sound that plays is of the block underneath. This PR adds these blocks to the proper tag to fix it, akin to carpets.

Similar to https://github.com/MehVahdJukaar/Supplementaries/pull/1922 .

The tag also exists in 1.20 so this can be cherrypicked.

### DIFF
--- a/src/generated/resources/data/minecraft/tags/block/combination_step_sound_blocks.json
+++ b/src/generated/resources/data/minecraft/tags/block/combination_step_sound_blocks.json
@@ -1,0 +1,7 @@
+{
+  "values": [
+    "farmersdelight:canvas_rug",
+    "farmersdelight:full_tatami_mat",
+    "farmersdelight:half_tatami_mat"
+  ]
+}


Mickeon:
It dawned on me that these tags are generated from some pre-compilation code. My work from web Github is limited, but these tags would have to be added there, instead.

vectorwing:
Thanks for the PR, and for letting me know such a tag exists!

Yes, direct edits to `generated` json's are not correct, as they will be erased on the next `runData`. That said, this tagging is easy enough to do on a quick commit, so I can get to it from my end as well.

In cases such as this, where you wish to propose useful tagging without editing datagen code, feel free to open a **discussion** post explaining the tag's purpose, or an **issue** if the lack of tagging causes a bug. 👍

vectorwing:
Added on commit 7045142, with credit on the changelog. I'll be closing this PR. Thank you again! 👍

PR #1247, opened 2026-03-06
fix: language fatal error

teaSummer:
It should be ported to other branches.


vectorwing:
Well spotted! Thank you. 👍

PR #1265, opened 2026-03-28
feat: abstract stove

Lordfirespeed:


Lordfirespeed:
If you'd prefer I can rebase onto the 1.21 branch and target that. Let me know.

Edit: I have also rebased onto 1.21, I can cherry-pick any revisions from here onto there and it should be ready to go.

Lordfirespeed:
I've only just noticed that indentation is buggered.

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
```diff
@@ -0,0 +1,219 @@
...
+		if (heldStack.canPerformAction(ToolActions.SHOVEL_DIG)) {
+			if (!level.isClientSide()) {
+				level.levelEvent(null, 1009, pos, 0);
+			}
+			extinguish(player, level, pos, state);
+			heldStack.hurtAndBreak(1, player, (action) -> action.broadcastBreakEvent(hand));
+			return InteractionResult.sidedSuccess(level.isClientSide());
+		}
+
+		if (heldStack.is(ForgeTags.BUCKETS_WATER)) {
```
vectorwing: The tag reference was updated in 1.3. It should be:
`CommonTags.Items.BUCKETS_WATER`

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
```diff
@@ -0,0 +1,219 @@
...
+			.setValue(LIT, true);
+	}
+
+	@Override
+	public void stepOn(Level level, BlockPos pos, BlockState state, Entity entity) {
+		burnEntitySteppingOnStove(level, pos, state, entity);
+		super.stepOn(level, pos, state, entity);
+	}
+
+	protected void burnEntitySteppingOnStove(Level level, BlockPos pos, BlockState state, Entity entity) {
```
vectorwing, line 178: Not sure if this method needs to be separate, as it contains almost all of the code that would go into `stepOn`.

src/main/java/vectorwing/farmersdelight/common/block/StoveBlock.java
```diff
@@ -3,228 +3,55 @@
...
-import vectorwing.farmersdelight.common.registry.ModDamageTypes;
 import vectorwing.farmersdelight.common.registry.ModSounds;
-import vectorwing.farmersdelight.common.tag.CommonTags;
-import vectorwing.farmersdelight.common.utility.ItemUtils;
-import vectorwing.farmersdelight.common.utility.MathUtils;
 
 import javax.annotation.Nullable;
-import java.util.Optional;
 
 @SuppressWarnings("deprecation")
```
vectorwing: The methods this annotation suppressed warnings for were moved to AbstractStoveBlock, so it should probably be moved there as well.

src/main/java/vectorwing/farmersdelight/common/block/entity/StoveBlockEntity.java
```diff
@@ -3,213 +3,64 @@
...
-		if (0 <= slot && slot < inventory.getSlots()) {
-			ItemStack slotStack = inventory.getStackInSlot(slot);
-			if (slotStack.isEmpty()) {
-				cookingTimesTotal[slot] = recipe.getCookingTime();
-				cookingTimes[slot] = 0;
-				inventory.setStackInSlot(slot, itemStackIn.split(1));
-				lastRecipeIDs[slot] = recipe.getId();
-				inventoryChanged();
-				return true;
+	private void addSmokeParticles() {
```
vectorwing: Wouldn't it be better if this was set to public, or provided in AbstractStoveBlockEntity?

I would imagine most stoves would just use the default smoke particles when grilling something, and this seems to lock that, requiring a copy/paste.

src/main/java/vectorwing/farmersdelight/common/block/entity/StoveBlockEntity.java
```diff
@@ -3,213 +3,64 @@
...
+		assert this.level != null;
```
vectorwing, line 25: `particleTick` has access to the client level, so passing it to this method would eliminate the need for this `assert`. Also, IIRC, `assert` actually throws an exception when it's null; perhaps just returning early would be best in such cases.

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
```diff
@@ -0,0 +1,215 @@
...
+import net.minecraft.world.phys.Vec2;
+import vectorwing.farmersdelight.common.block.AbstractStoveBlock;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.stream.IntStream;
+
+public abstract class AbstractStoveBlockEntity extends BlockEntity implements Clearable {
+	private final NonNullList<ItemStack> items;
+	private final RecipeType<? extends AbstractCookingRecipe> recipeType;
```
vectorwing: The `recipeType` field is unused, as all of its uses seem to happen inside the constructor.

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
```diff
@@ -0,0 +1,215 @@
...
+
+		if (state.getValue(AbstractStoveBlock.LIT)) {
+			stoveEntity.cookAndOutputItems();
+		} else {
+			stoveEntity.coolItems();
+		}
+	}
+
+	private void cookAndOutputItems() {
+		assert this.level != null;
```
vectorwing, line 114: `assert` should be avoided in production code, as it throws exceptions on fail.
Here, it may be better to just return when null instead.

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
```diff
@@ -0,0 +1,215 @@
...
+import net.minecraft.world.level.gameevent.GameEvent;
+import net.minecraft.world.phys.Vec2;
+import vectorwing.farmersdelight.common.block.AbstractStoveBlock;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.stream.IntStream;
+
+public abstract class AbstractStoveBlockEntity extends BlockEntity implements Clearable {
+	private final NonNullList<ItemStack> items;
```
vectorwing: Is there a reason why you moved away from using an ItemStackHandler here?

I recall updating all FD workstations to make use of this system years ago, and it did provide more control over stack handling in most cases. Not sure if there's a need to move away from it again.

vectorwing (changes requested):
Alright, I did a first pass over the code, and found a few inconsistencies.

The code works when testing locally, and I can see you fixed the issue where the Stove won't spill its contents when blocked while unlit, which is good.

I can do a second pass after the current comments are addressed. 👍

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
Lordfirespeed: Will do 👍🏻

src/main/java/vectorwing/farmersdelight/common/block/StoveBlock.java
Lordfirespeed: Good point

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
Lordfirespeed: In the original version of the code, the `base.stepOn()` was invoked at the end of the method, but I wanted to use early-return pattern to make this method easier to read.

To keep the order of invocation the same and use early return pattern, I had to extract this method

src/main/java/vectorwing/farmersdelight/common/block/entity/StoveBlockEntity.java
Lordfirespeed: Agree, this should be public or protected

src/main/java/vectorwing/farmersdelight/common/block/entity/StoveBlockEntity.java
Lordfirespeed: assertions are not enabled by default

the assertion is here to make it clear to developers reading the code that the method must only be invoked in contexts where the level is definitely not null

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
Lordfirespeed: I included the field to mirror the AbstractFurnace class, which also retains a similar field, though I don't remember whether I checked for usages of that 🤔 

I thought it was a good idea to keep the field around 'just in case'

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
Lordfirespeed: Assertions are not enabled by default
Edit: I verified this, `assert false` didn't throw any exception at runtime
of course, passing a `null` value to the `level` parameter of whatever function needs it would probably cause an exception, but the point of the assertion is to ensure that `this.level` is actually never `null` because developers know they must only call the function when `this.level` is definitely not null

As previous: this is primarily to signal to developers reading the code that the method must be invoked only when the level is definitely not null

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
Lordfirespeed: I wasn't sure whether it was intentional to use the ItemStackHandler or whether it was left over from something old - the campfire uses an item list and I was matching that.

I can see the appeal of the ItemStackHandler for blocks with GUIs but I thought the item list was more intuitive and yielded more elegant code

I'm happy to revert to the ItemStackHandler if that's your preference

vectorwing: Yeah, I'd rather use what was already in place. Even if the Stove isn't extending it, the Cooking Pot extends it for more specific control over the slots, so the consistency works better across them.

src/main/java/vectorwing/farmersdelight/common/block/entity/AbstractStoveBlockEntity.java
vectorwing: It seems AbstractFurnaceBlockEntity uses `recipeType` to pass it to the `ForgeHooks.getBurnTime()` method, which is mostly unused as well... I assume it was to make different furnaces burn fuel for different lengths of time, but doesn't seem to be the case.

Since the Stove doesn't use fuel, and furnaces don't seem to do much with it either, I assume it's safe to remove.

vectorwing:
I added a tiny change to the Stove, where it makes a sound when placing items on the grill. That might cause a small conflict, sorry about that. 😅

Lordfirespeed:
Ready for second pass

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
```diff
@@ -0,0 +1,220 @@
...
+
+		return InteractionResult.PASS;
+	}
+
+	protected InteractionResult tryToExtinguish(BlockState state, Level level, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit) {
+		ItemStack heldStack = player.getItemInHand(hand);
+
+		if (heldStack.canPerformAction(ToolActions.SHOVEL_DIG)) {
+			if (!level.isClientSide()) {
+				level.levelEvent(null, 1009, pos, 0);
```
vectorwing: For level events, it would be more readable if we used the LevelEvent static fields directly, so that others reading the code know that 1009 is `SOUND_EXTINGUISH_FIRE`.

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
```diff
@@ -0,0 +1,220 @@
...
+	public void extinguish(@Nullable Entity entity, LevelAccessor level, BlockPos pos, BlockState state) {
+		if (level.getBlockEntity(pos) instanceof AbstractStoveBlockEntity stoveEntity) {
+			stoveEntity.extinguish();
+		}
+		level.gameEvent(entity, GameEvent.BLOCK_CHANGE, pos);
+		if (level.isClientSide()) return;
+
+		var newState = state.setValue(LIT, false);
+		level.setBlock(pos, newState, 11);
+		level.gameEvent(GameEvent.BLOCK_CHANGE, pos, GameEvent.Context.of(entity, newState));
```
vectorwing: Question: why is this game event being sent twice?

vectorwing (commented):
Second pass done. Aside from a comment and a quick question, everything here seems in order. You managed to organize one of the oldest bits of code in FD, it's looking pretty tidy! 👍

I think we're ready for a merge. I can address the comment below on my own later, unless you'd like to.

src/main/java/vectorwing/farmersdelight/common/block/AbstractStoveBlock.java
```diff
@@ -0,0 +1,220 @@
...
+			if (extinguishResult != InteractionResult.PASS) return extinguishResult;
+		} else {
+			var igniteResult = tryToIgnite(state, level, pos, player, hand, hit);
+			if (igniteResult != InteractionResult.PASS) return igniteResult;
+		}
+
+		return tryToPlaceFoodItem(state, level, pos, player, hand, hit);
+	}
+
+	protected InteractionResult tryToIgnite(BlockState state, Level level, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit) {
```
vectorwing, line 75: Igniting currently lacks the gameEvent call, so it doesn't trigger sculk sensors. Might be worth adding here too.

PR #1279, opened 2026-04-25
Update translations

BoredYukolin:
Updated translations for zh_cn, zh_tw, zh_hk and lzh.

Please apply the changes to all updating branches as well :)

### DIFF
--- a/src/main/java/vectorwing/farmersdelight/common/tag/ModTags.java
+++ b/src/main/java/vectorwing/farmersdelight/common/tag/ModTags.java
@@ -139,7 +139,7 @@ public class ModTags
 	}
 
 	/**
-	 * The following tag references are DEPRECATED, and should no longer be used by add-ons. Please use the sub-classed tag references moving forward.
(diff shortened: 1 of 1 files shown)


vectorwing:
Thanks for the lang update!

Could you please point this PR to the `dev/1.3` branch? So it can be included in the next update.

vectorwing:
Actually, nevermind. The update will release soon, so I can just add them before release.

src/main/resources/assets/farmersdelight/lang/zh_tw.json
```diff
@@ -1,8 +1,7 @@
 {
   "block.farmersdelight.stove": "爐灶",
   "block.farmersdelight.cooking_pot": "烹飪鍋",
-  "block.farmersdelight.wooden_basket": "木籃",
```
vectorwing, line 4: These two lines are correct, as they are a part of the next update. `basket` is now `wooden_basket` and `bamboo_basket`.

It might be worthwhile to remake this PR into the `dev/1.3` branch to catch the new langs.

vectorwing (changes requested):

PR #1300, opened 2026-04-30
Updated Korean Translation

Kardane:
Nice Update!
It's been long time with no translation update, but here we go!


vectorwing:
Thanks for the update!

PR #1319, opened 2026-05-11
Feat: Add Bulgarian translation

OmegaSleepy:
Me and a few friends (all Bulgarian natives) really like Farmers Delight and wanted to contribute to the project by adding translations about all String that the mod has. 
If the maintainers like our translations, we would be glad to continue translating the mod in the future to our native language :)


vectorwing:
Thank you for the brand new lang! ^^
I'll credit it as "OmegaSleepy and friends" in the changelog. 👍

PR #1347, opened 2026-06-05
26.1 porting contributions

AyOhEe:
As yet still in progress, but I've done the following:
- Replace MethodsReturnNonnullByDefault annotations (moved from `net.minecraft` to `com.mojang.logging.annotations`)
- FDBlockLoot now compiles
- (Hanging)Canvas sign rendering and editing, and associated ModAtlas changes
- `KeepRichSoilUntrampledMixin` fixed (`FarmBlock` -> `FarmlandBlock`)
- `FlipSkilletPayload` is now sent correctly
- `SkilletRenderer`, `DefaultStoveRenderer`, and `CuttingBoardRenderer` have been moved to the new rendering system
- Cooking and Cutting recipes compile, and have proper serializers

Some of this is very much still in the works and temporary just so it'll compile. Particularly, canvas sign rendering could be a lot less complex and duplicated if I had used an access widener. Apparently transactions are also `AutoCloseable`, so that side of things could be much cleaner.


vectorwing:
Thanks for the assistance! ^^

I'm currently taking a break from the port, but I might be able to give this a look next week. I'll post a quick review once I'm able.

AyOhEe:
Glad to be of help! I’m a bit unfamiliar with the 26.1 ways of doing
things, so at least some of this will probably be a bit shaky. I should
have it compiling by the time you can review. I think it’s just block/item
models, a bit of reworking of the cooking pot, and moving the skillet from
a BEWLR to special model rendering.

src/main/java/vectorwing/farmersdelight/common/event/CommonEvents.java
```diff
@@ -62,6 +64,12 @@ public void onServerWorldTick(ServerTickEvent.Post event) {
 		}
 	}
 
+	// TODO this absolutely sucks but might be necessary for JEI recipe registration
+	@SubscribeEvent
+	public void onServerAboutToStart(ServerAboutToStartEvent event) {
+		FDRecipes.SERVER = event.getServer();
```
cassiancc, line 70: This is the wrong approach and will break on multiplayer servers. Recipes should be sent via `OnDatapackSyncEvent` in the common event handler and retrieved in the client event handler in `RecipesRecievedEvent`. 
https://docs.neoforged.net/docs/resources/server/recipes/#client-side-recipes

cassiancc (changes requested):

src/main/java/vectorwing/farmersdelight/common/event/CommonEvents.java
AyOhEe: Missed that entirely in the docs, thanks for flagging that. My poor assumption was that JEI was syncing provided recipes when installed server-side. Will replace with correct approach.

cassiancc: In theory it could sync recipes itself, but there's no guarantee every recipe type is safe to synchronize, so it's left as an opt-in system.

PR #1381, opened 2026-08-28
Unhardcode mushroom colony transformations

GizmoTheMoonPig:
Hi!
I'm working on an addon for the mod and when attempting to make my own mushroom colony, I noticed that colony transformations are unfortunately hardcoded in the rich soil block. In order to have other mushrooms transform into colonies, you need a mixin. I like to avoid mixins as much as possible and figured a proper system for adding new transformations was needed. This PR adds a new DataMap that allows people to define what blocks should transform into others when growing on rich soil. As this is a datamap, you can technically use any block you want in this, its not limited to just mushrooms. I figured this was an elegant enough solution, but if you have a better idea, please let me know. Thanks!


vectorwing:
Thanks for the PR! But I already implemented this feature on the `update-1.21/1.4` branch! 😅

It was one of the first things I added to the upcoming update, now that 1.20.1 has been dropped. It may take a while to come out, but it should definitely help out for add-ons. 👍

GizmoTheMoonPig:
oh sweet, no worries then. Thanks for taking care of that :)

